PR #1, opened 2016-12-15
Added ownership, fees, setPrice(), setFee and correct decimals.

AugustoL:
Added ownership, fees gathered on every transfer with value, setPrice and setFee methods added, using correct decimals, better and more test cases checking every field of the contract.

The token contract now will have an owner, can be one person or an entire organization, it will gain money in tokens charging 1% fee per transaction with value.

+ Test cases, checking that the main values are the correct on every test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Lif Token is based on the ERC20 token protocol, with the option to also send inf
 ```sh
 npm install ethereumjs-testrpc -g
 npm install truffle -g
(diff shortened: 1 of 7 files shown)


kvbik:
as far as i understand it right now, let's go 👍  :)))

AugustoL:
@kvakes the price can be changed by the owner of the contract, the owner can be a single person or organization like a DAO for example.

PR #9, opened 2017-04-13
Install truffle and ethereumjs-testrpc as devDependencies

nberger:
Makes installation easier and uses explicit version numbers


kvbik:
👍 btw i will push my docker settings too soon.

PR #13, opened 2017-04-18
Run tests with a single command

nberger:
Takes care of running testrpc first, there's no need to have it running separately now

Extracted from code in #10 by @AugustoL :)


kvbik:
👍 i will just push some of my docker setups too, so we are on the same page.

PR #19, opened 2017-04-23
Added call execution on transfers, and fix bug ot send to token.

### LINKED ISSUE #14: Add option to call data between contracts.
The idea is to change the transferDataFrom and transferData on the LifToken contract to also call the data between contracts using a Lif Token transference, for this the data parameter need to be changed to bytes and add a boolean as parameter that if its true it will execute the call.

Something like this:
```
// ERC20 transfer method but with data parameter.
    function transferData(address to, uint value, bytes data, bool makeCall) onlyTokenHolder() onStatus(3,4) returns (bool success) {

      // If transfer have value process it
      if (value > 0) {
        balances[msg.sender] = safeSub(balances[msg.sender], value);
        balances[to] = safeAdd(balances[to], value);
        giveVotes(msg.sender, to);
      }

      if (makeCall && to.call(data)
        TransferData(msg.sender, to, value, data);
      else
        TransferData(msg.sender, to, value, data);

      return true;

    }
```

AugustoL:
Added call execution on transferData calls, not allowing transfers and approvals to token contract. Fix #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,13 +43,13 @@ npm install
 
 ## New Token Methods
 
-Líf token is ERC20 compatible but it also has two more methods to allow the transference of data between users/contracts.
+Líf token is ERC20 compatible but it also has two more methods to allow the transference of data and execution of calls between users/contracts.
 
 ### transferData
 
 Transfer tokens from one address to another.
 ```
-transferData(address _to, uint _value, string _data) => (bool success)
+transferData(address to, uint value, bytes data, bool doCall) => (bool success)
 ```
 Returns: bool, Success of the operation.
 
@@ -57,7 +57,7 @@ Returns: bool, Success of the operation.
 
 Transfer  an allowed amount of tokens from one address to another.
 ```
-transferDataFrom(address _from, address _to, uint _value, string _data) => (bool success)
+transferDataFrom(address from, address to, uint value, bytes data, bool doCall) => (bool success)
 ```
 Returns: bool, Success of the operation.
 
@@ -75,22 +75,22 @@ This functions can be called only internally, in order to do that a token holder
 ```
 setBaseProposalFee(uint _baseProposalFee)
 setProposalBlocksWait(uint _proposalBlocksWait)
-sendEther(address _to, uint _amount)
-setStatus(uint _newStatus)
-addDAOAction(address _target, uint _balanceNeeded, bytes4 _signature)
+sendEther(address to, uint amount)
+setStatus(uint newStatus)
+addDAOAction(address target, uint balanceNeeded, bytes4 signature)
 ```
 
 ### New Proposal
 
 Creates a new proposal on the token, the token holder that creates the proposal needs to have the more than the fee charged for the proposal creation.
 ```
-newProposal(address _target, uint _value, string _description, uint _agePerBlock, bytes4 _signature, bytes _actionData)
+newProposal(address target, uint value, string description, uint agePerBlock, bytes4 signature, bytes actionData)
 ```
 ### Vote
 
 Vote a proposal for yes or no using the proposal ID.
 ```
-vote(uint _proposalID, bool _vote)
+vote(uint proposalID, bool vote)
 ```
 
 ## Test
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -43,21 +43,21 @@ npm install
 
 ## New Token Methods
 
-Líf token is ERC20 compatible but it also has two more methods to allow the transference of data between users/contracts.
+Líf token is ERC20 compatible but it also has two more methods to allow the transference of data and execution of calls between users/contracts.
 
 ### transferData
 
 Transfer tokens from one address to another.
```
nberger: The method name is `transferData` but it says it's used to transfer tokens. Is that correct? Is it because it actually transfer both tokens and data?

README.md
```diff
@@ -43,21 +43,21 @@ npm install
...
 ```
-transferData(address _to, uint _value, string _data) => (bool success)
+transferData(address to, uint value, bytes data, bool doCall) => (bool success)
```
nberger: What's the semantic difference of both values of `doCall`?

I mean, when it's false it does it mean that data is just data, but when it's true it means that data has all the information to execute a method in the contract (I guess it has the method name and the args)? If so, I think we could make the api more clear by separating this into 2 separate methods, one to transfer data and another for method calls. Or at least explain a bit more about `doCall`

contracts/LifToken.sol
```diff
@@ -457,14 +469,20 @@ contract LifToken is Ownable, ERC20, SafeMath, PullPayment {
...
 
       return true;
 
     }
 
     // ERC20 transferFrom method but with data parameter.
-    function transferDataFrom(address from, address to, uint value, string data) onStatus(3,4) returns (bool success) {
+    function transferDataFrom(address from, address to, uint value, bytes data, bool doCall) onStatus(3,4) returns (bool success) {
+
+      if (to == address(this))
```
nberger: should we extract this check to a fn like `checkNotSelfTransfer`? I'm ok if it's in a further PR

contracts/LifToken.sol
```diff
@@ -457,14 +469,20 @@ contract LifToken is Ownable, ERC20, SafeMath, PullPayment {
         giveVotes(msg.sender, to);
       }
 
-      TransferData(msg.sender, to, value, data);
+      if (doCall && to.call(data))
+        TransferData(msg.sender, to, value, data);
+      else if (!doCall)
```
nberger: So when `!(doCall && to.call(data))` it seems like we just return true, is that correct? Like it fails silently, or do we just expect `to.call(data)` to throw when it fails?

README.md
AugustoL: if doCall is true it will execute the bytes that are sent on a call, if not it will just trigger the event, so this function can be used to register something on the blockchain, mostly signatures, or execute calls between token holders, we are near the gas limit on contract deployment so adding another function to do almost the same I think is not possible.
And yes, we will need to explain a bit more about what doCall can do.

contracts/LifToken.sol
AugustoL: Yes, we can create a modifier like checkNotSelfTransfer , future PR :)

README.md
AugustoL: Right, you can transfer only data, or both.

contracts/LifToken.sol
AugustoL: We expect the call to always return a bool, so if doCall is false it will just jump to the next condition, where it will just log the event

README.md
nberger: > we are near the gas limit on contract deployment

Interesting. How does this work? I mean I read a bit about the gas cost and limits, I have a rough idea of what you wanted to mean, but how do you near that we are near the limit? How do you measure it?

AugustoL: Because I remember a week ago when I was working on the contract I could deploy the contract because gasLimit and I had to stop adding new things and do a refactor of the code to save space, maybe now we cant, but it I think is calculated by: CONTRACT_DATA_BYTES * GAS_PRICE, this has to be less than GAS_LIMIT, with is now near 4 million wei. You can now how much it cost seeing the gasUsed on a contract deployment tx.
In a near future zeppelin libs are planned to be hosted on the blockchain, and that will help up save space.
Although I been thinking on develop another smart contract to just do the crowdsale stage, this can also be a solution if we need to save gas in a future.

nberger: Cool, thanks for the info, learning much over here :)

PR #26, opened 2017-05-04
[WIP] Add stop & resume transactions

### LINKED ISSUE #18: Add stop in emergency test case
In case of a emergency the token owner should be able to stop all contract operations changing contract status.

nberger:
Fixes #18


test/LifToken.js
```diff
@@ -32,6 +32,24 @@ contract('LifToken', function(accounts) {
...
 
+  it("should be able to stop and resume the token", async function() {
+    await token.stop();
+
+    await help.addCrowdsaleStage(token, 10000000, price)[1];
+
+    var price = web3.toWei(0.1, 'ether');
+
+    assert.equal(await token.submitBid(accounts[1], 1000, { value: 1000*price, from: accounts[1]}),
+      false, // TODO: this doesn't work of course, how can I assert that the tx failed?
```
nberger: @AugustoL: I don't know how to assert that the transaction can't go through because of the `onStatus(2,3)`. I know that this is not the way, I wonder what the way is. Any hint would be appreciated :)

AugustoL: @nberger adding another uint and function on the token contract is not necessary, you can use setStatus() function.
The setStatus() function can be called by the owner or DAO, the idea is to test it on both ways.
https://github.com/windingtree/LifToken/blob/master/contracts/LifToken.sol#L409

https://github.com/windingtree/LifToken/blob/master/test/DAO.js#L175 You can use this test to test the setStatus form DAO proposal

AugustoL:
Also, I noted that there is a from fromDAO() modifier that is not needed, please remove it :)

nberger:
This is far from complete and way behind master so I'll close it now

PR #31, opened 2017-05-10
Added approveData, using abiDecoder and fix zeppelin version.

### LINKED ISSUE #29: Use tx.origin instead of msg.sender on data transactions
The methods transferData and transferDataFrom was designed with the purpose to execute function between contracts that are part of the Winding Tree platform, because of this the contract has to refer tx.origin as sender of the function and balances instead of msg.sender, if not it can lost track of the sender address, and since this functions dont belong to the ERC20 interface we can apply that change.

### LINKED ISSUE #30: Add approveData data function
We need a to be able to execute a calla and approve tokens spending on a unique transaction, for this we need the approve function of ERC20 but with the bytes and bool to execute call, right now we have only the methods of transfer and transferFrom, but if we want to make a booking of an inventory and show the owner of the items that we have that balance to make the payment at his disposal we need this method.

```
function approveData(address spender, uint value, bytes data, bool doCall) onStatus(3,4) returns (bool success) {

      if (spender == address(this))
        throw;

      allowed[tx.origin][spender] = value;

      if (doCall && spender.call(data))
        ApprovalData(tx.origin, spender, value, data);
      else if (!doCall)
        ApprovalData(tx.origin, spender, value, data);

      return true;

    }
```

AugustoL:
Fix #30 Fix #29 

Using zeppelin-solidity 1.0.4 version because next version use safeMath as lib and break our tests.

### DIFF
--- a/contracts/LifToken.sol
+++ b/contracts/LifToken.sol
@@ -128,6 +128,7 @@ contract LifToken is Ownable, ERC20, SafeMath, PullPayment {
 
     // Edit of the ERC20 token events to support data argument
     event TransferData(address indexed from, address indexed to, uint value, bytes data);
+    event ApprovalData(address indexed from, address indexed spender, uint value, bytes data);
 
     // Proposal events
     event proposalAdded(uint proposalId);
@@ -455,7 +456,24 @@ contract LifToken is Ownable, ERC20, SafeMath, PullPayment {
 
     }
 
-    // ERC20 transfer method but with data parameter.
+    //ERC20 token approve method with data call/log option.
+    function approveData(address spender, uint value, bytes data, bool doCall) onStatus(3,4) returns (bool success) {
+
+      if (spender == address(this))
+        throw;
+
+      allowed[tx.origin][spender] = value;
+
+      if (doCall && spender.call(data))
+        ApprovalData(tx.origin, spender, value, data);
+      else if (!doCall)
+        ApprovalData(tx.origin, spender, value, data);
+
+      return true;
+
+    }
+
+    // ERC20 transfer method with data call/log option.
     function transferData(address to, uint value, bytes data, bool doCall) external onStatus(3,4) returns (bool success) {
 
       if (to == address(this))
@@ -463,21 +481,21 @@ contract LifToken is Ownable, ERC20, SafeMath, PullPayment {
 
       // If transfer have value process it
       if (value > 0) {
-        balances[msg.sender] = safeSub(balances[msg.sender], value);
+        balances[tx.origin] = safeSub(balances[tx.origin], value);
         balances[to] = safeAdd(balances[to], value);
-        giveVotes(msg.sender, to);
+        giveVotes(tx.origin, to);
       }
 
       if (doCall && to.call(data))
-        TransferData(msg.sender, to, value, data);
+        TransferData(tx.origin, to, value, data);
       else if (!doCall)
-        TransferData(msg.sender, to, value, data);
+        TransferData(tx.origin, to, value, data);
 
       return true;
 
     }
 
-    // ERC20 transferFrom method but with data parameter.
+    // ERC20 transferFrom method with data call/log option.
     function transferDataFrom(address from, address to, uint value, bytes data, bool doCall) external onStatus(3,4) returns (bool success) {
 
       if (to == address(this))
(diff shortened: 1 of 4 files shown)


test/LifToken.js
```diff
@@ -79,9 +79,10 @@ contract('LifToken', function(accounts) {
 
     var dataEncoded = help.hexEncode(JSON.stringify({awesomeField:"AwesomeString"}));
     let transaction = await token.transferData(accounts[2], help.formatBalance(1000), dataEncoded, false, {from: accounts[1]});
+    let decodedEvents = help.abiDecoder.decodeLogs(transaction.receipt.logs);
```
nberger: I'll try to learn more on my own, but could you explain briefly what's the difference between `transaction.logs` and `transaction.receipt.logs`?

test/LifToken.js
```diff
@@ -141,26 +146,43 @@ contract('LifToken', function(accounts) {
...
+    assert.equal(data, decodedEvents[1].events[3].value);
+    assert.equal('0x1e24000000000000000000000000000000000000000000000000000000000000', decodedEvents[0].events[0].value);
+    assert.equal(666, decodedEvents[0].events[1].value);
+    assert.equal('Transfer Done', decodedEvents[0].events[2].value);
+    assert.equal(help.formatBalance(1000), await token.balanceOf(message.contract.address));
 
     await help.checkValues(token, accounts,1000000, 10000000, 0, [3999000,3000000,2000000,1000000,0]);
   });
 
+  it("should return correct balances after approve and show the event on receiver contract", async function() {
```
nberger: "correct balances" doesn't help me understand what's being asserted in the test. Maybe something like "unmodified balances after approve"? To make it more clear that they are not modified, should we have the same `checkValues` assertion before and after the call to approve?

package.json
```diff
@@ -20,11 +20,12 @@
   "author": "Augusto Lemble <me@augustolemble.com>",
   "license": "ISC",
   "dependencies": {
+    "abi-decoder": "^1.0.8",
     "ethereumjs-abi": "^0.6.4",
     "protobufjs": "^6.0.2",
     "solc": "^0.4.9",
     "truffle-hdwallet-provider": "0.0.3",
-    "zeppelin-solidity": "^1.0.4"
+    "zeppelin-solidity": "1.0.4"
```
AugustoL: I think we need a separated PR if we upgrade zeppelin-solidity, but not to continue using the same version. And we will need a new PR to update zeppelin solidity.

contracts/LifToken.sol
```diff
@@ -455,47 +456,64 @@ contract LifToken is Ownable, ERC20, SafeMath, PullPayment {
 
     }
 
-    // ERC20 transfer method but with data parameter.
+    //ERC20 token approve method with data call/log option.
+    function approveData(address spender, uint value, bytes data, bool doCall) onStatus(3,4) returns (bool success) {
```
AugustoL: This is defined on the ERC20 token standard, its one of the few things that we shouldnt change to follow the standard, if we were on another function we can. And the spender not necessarily needs to eb a token holder, it can be an address without any tokens too :).

nberger: Oh right, so the spender is the one who spends the token from some other token holder. So now the name makes more sense ;) :+1:

test/LifToken.js
AugustoL: We execute calls inside our contracts, this are internal calls that can trigger events too. The events on the first transaction and function called are on transaction.logs, and the events triggered on internal functions are on transaction.receipt.logs. So if you want to get all the events you should look at transaction.receipt.logs, the bad thing is that here the events are encoded, but we decode them using abi-decoder lib

package.json
nberger: The way I see it, they are unrelated changes, and I like to use separate PRs for unrelated changes, except when there's a good reason to keep them together. Thinking that way, we could even have 3 separate PRs here: the change to `tx.origin`, fixing the zeppelin-solidity version, adding `approveData`.

test/LifToken.js
nberger: Nice, got it, thanks :)

test/LifToken.js
AugustoL: For me is the same, I mean, the correct balances in this case would be that the balances stay the same haha. We have already a checkValues after approve and before approve we know that the balances are correct because there is one test case that check the simulateCrowdsale works correctly.

package.json
AugustoL: @nberger if you want to separate it in different PRs you can do that, but if not we can just go ahead and merge this PR. Your choice, but if you want to make it in 3 PRs you will need to do it hahaha.

test/LifToken.js
nberger: Hehe. The thing is that you know very precisely what `simulateCrowdsale` does, including what the balances are on each account, without having to go and read the code. Someone new (or relatively new) to the codebase doesn't know that, and it would help him to understand what approveData is supposed to do. Remember that the tests can also act as documentation, and "correct balances" doesn't say much when trying to understand what's happening. It's like we could add "it should have correct balances" to each and every test ;)

package.json
nberger: @AugustoL no, it's ok, let's keep it as is. in this case, they are small changes, so they don't even make the review much more difficult. I wanted to mention it thinking in future PRs, and also to have this discussion :)

test/LifToken.js
AugustoL: :+1: Yes, you are right, well, we need to do a review of all test cases to make them more understandable , future task.

PR #33, opened 2017-05-11
Add missing totalSupply state var. Tweak comments

nberger:


AugustoL:
@nberger The totalSupply is inherited form the ERC20 contract, but on https://github.com/windingtree/LifToken/pull/32 is removed and added to the contract because ERC20 inheritance will cause conflicts. So that change is already done on that PR

nberger:
Got it, thanks @AugustoL!

PR #42, opened 2017-06-13
Extract LifCrowdsale contract

### LINKED ISSUE #34: Extract the crowdsale from LifToken into a separate contract
The goal is to decouple the token core features from the crowdsale, it will make LifToken more simple and easier to reason about. We can think of the Crowdsale as a contract that holds LifTokens and distributes them when it's finished, while also giving the non-distributed tokens back to the LifToken. Looking at it that way, the `LifCrowdsale` lifecycle could be like the following:

- Create LifCrowdsale (by owner)
- Transfer tokens to LifCrowdsale (by owner)
- Submit Bids (by anyone who wants to buy LifToken)
- Distribute tokens (called by owner or by buyer, uses `transferFrom`?)
- Give back not-distributed tokens to LifToken (by owner)

nberger:
Fixes #34 

TODO:
- [ ] Find a way for the owner to be able to reuse unused tokens from a Crowdsale (for example to fund a new Crowdsale)
- [x] Fix tests (mostly `simulateCrowdsale` but other minor stuff too)

### DIFF
--- a/contracts/FuturePayment.sol
+++ b/contracts/FuturePayment.sol
@@ -1,0 +1,33 @@
+pragma solidity ^0.4.8;
+
+import './zeppelin/token/ERC20Basic.sol';
+
+contract FuturePayment {
+
+    // Represents a token payment that can be claimed after certain block from an address
+    address public payee;
+    uint public afterBlock;
+    address public tokenAddress;
+
+    event PaymentDone();
+
+    function FuturePayment(address _payee, uint _afterBlock, address _tokenAddress) {
+        payee = _payee;
+        afterBlock = _afterBlock;
+        tokenAddress = _tokenAddress;   
+    }
+
+    // Function that allows an address to claim a futurePayment on tokens
+    function claimPayment() external {
+
+      if ((payee != msg.sender) || (afterBlock > block.number))
+        throw;
+
+      ERC20Basic token = ERC20Basic(tokenAddress);
+
+      token.transfer(payee, token.balanceOf(address(this)));
+
+      PaymentDone();
+    }
+
+}
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -241,9 +241,9 @@ contract LifCrowdsale is Ownable, PullPayment {
         uint presaleTokens = getPresaleTokens(lastPrice);
 
         if (ownerPercentage > 0) {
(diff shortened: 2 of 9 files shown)


nberger:
@AugustoL: an important TODO here is to find a way for the owner to be able to fund a new crowdsale from the unused tokens resulting from the difference between the maxTokens and the soldTokens in a finished crowdsale. The current implementation is returning those unused tokens to the LifToken so now those tokens would be in the balance of the token itself. The thing is that the owner does not have allowance to move those tokens, so in the practice they are frozen there.

One way to circumvent this frozen tokens issue would be to allow the owner to transfer any tokens from the LifToken balance but only while the `issuedTokens <= OWNER_SUPPLY`.

But you mentioned another solution a while ago which is to destroy the tokens instead of returning them to the LifToken. So any token holder would be able to `destroyTokens` that she owns, and that includes the unused tokens from a Crowdsale. Once they are destroyed, the owner would be able to issue again to fund a new crowdsale.

What do you think? Do you see any downside or security issue in adding the destroyTokens transaction?

AugustoL:
@nberger I think that the best solution will be to return the remaining tokens to the Lif contract and allow the owner to send them from there. BUT taking in mind that the owner will have forever and ever total control of that first batch of tokens it would be the same and simpler to send the remaining tokens to the owner multisig contract. And we can create crowdsale contracts from that multisig contract. Doing this we wont need to add more code to the Lif token contract, and we can leave the control of ti to the token holders.

nberger:
> BUT taking in mind that the owner will have forever and ever total control of that first batch of tokens 

Is this exactly that way? I thought that the DAO could determine to give a different use to the tokens, so it's not really forever and ever but until the owner OR the DAO gives them a different usage. Even though the owner could take ownership or determine how to use them anytime before the DAO does it, it needs an active action from the owner if the tokens are on the LifToken or destroyed. If they sit in the owner multisig contract then the owner has exclusive power on those tokens.

AugustoL:
I think that should be the way, that the only way to the token god would be if you own the token contract and have power over a great amount of tokens, and votes too. WT will be able to gather certain percentage of the ETH raised by the the first 10M tokens sold. And organize the DAO at first, I imagine that after the crowdsale ends we will be responsible of mantaining the DAO contract but not owners on it, the power would be on the token holders. If we didnt selll all the tokens on the first crowdsale we still have them to decide when we continue distribute them.

PR #45, opened 2017-06-22
Upgrade to zeppelin 1.0.6

nberger:


AugustoL:
@nberger last stable version is 1.0.7

AugustoL:
Fixing the short address attack! Great @nberger !

nberger:
> @nberger last stable version is 1.0.7

@AugustoL: That was my first thought, but then the releases page in their github shows "1.0.6" as the latest release. What do you think if we merge this upgrade that includes the short address attack fix as you noticed :) and then we upgrade separately to 1.0.7?

selection_374

selection_373

AugustoL:
@nberger my bad, deal, lets do that :)

nberger:
Cool! Just fixed a syntax error. Let's see if tests run ok, then I'll squash a fix commits and merge :)

PR #52, opened 2017-07-03
Modify presale payments policy from discount to bonus rate

### LINKED ISSUE #47: Change discount rate rule to bonus rate for the Presale

nberger:
Fixes #47

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -37,7 +37,7 @@ contract LifCrowdsale is Ownable, PullPayment {
     uint public minCap;
     uint public maxCap;
     uint public totalTokens;
-    uint public presaleDiscount;
+    uint public presaleBonusRate;
     uint public ownerPercentage;
     uint public totalPresaleWei;
     uint public weiRaised;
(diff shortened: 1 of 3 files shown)


contracts/LifCrowdsale.sol
```diff
@@ -102,39 +102,44 @@ contract LifCrowdsale is Ownable, PullPayment {
...
+      // check that crowdsale balance is AT LEAST max(with bonus)tokens at lowest possible valuation +
       // founders tokens
       uint crowdsaleBalance = ERC20Basic(tokenAddress).balanceOf(address(this)).div(LONG_DECIMALS);
 
       if (crowdsaleBalance < getMaxTokens())
         throw;
     }
 
-    function distributeTokens(address buyer, bool discount) external onStatus(3, 0) {
+    function getBuyerPresaleTokens(address buyer) public returns (uint) {
```
nberger: What's the impact of having it as a separate function? I'd like to keep it this way because I think it makes the code more clear. And it's also called from the tests for debugging purposes (but we could also have assertions on it).

I'll try to measure the impact on gas consumption, I assume you suggest inlining it because of that but I'd like to understand how much additional gas we would consume and on which transactions (is it only on `distributeTokens` txs or also in others?)

AugustoL: The only impact is having more gas consumption on the deployment of this contract and the other contracts that use it, but if its not to much and the function helps debugging and it makes the contract more readable we should keep it.

nberger: From my tests using `web3.eth.getEstimate` I don't see substantial differences on gas consumption with having the function or not having it. Contract creation (measured with smth like `console.log("token creation gas estimate:", web3.eth.estimateGas({data: token.getData}));`) stays at 53000 in both cases. And the distributeTokens transaction has a minor extra consumption from 102021 to 102087 (that's an 0.064693% increase) which I think is worth it to pay if we prefer the code this way.

AugustoL: :+1: great, lets keep the function.

PR #56, opened 2017-07-06
Use yarn in travis

### LINKED ISSUE #55: Use yarn and cache dependencies in travis for faster builds

nberger:
Fixes #55


nberger:
It seems like it cuts down from ~4:40 to ~4:00 build time (~15% reduction)

PR #58, opened 2017-07-06
Remove short address checks: not effective and bring new problems

### LINKED ISSUE #57: Remove short address checks as they are not effective & can create problems
More information: https://blog.coinfabrik.com/smart-contract-short-address-attack-mitigation-failure/

TL;DR:

> On one hand, no one has found a way to take advantage of the Short Address Attack if the exchanges verify that the addresses’ format is correct. On the other hand, this mitigation caused problems for some smart contracts. For those reasons, we suggest removing the  onlyPayloadSize modifier and enforcing security checks on the exchange.

nberger:
More info: https://blog.coinfabrik.com/smart-contract-short-address-attack-mitigation-failure/

Fixes #57

### DIFF
--- a/contracts/LifToken.sol
+++ b/contracts/LifToken.sol
@@ -125,16 +125,6 @@ contract LifToken is Ownable, PullPayment {
         _;
     }
 
-    // Used to implement fix for short address attack. More information and original
-    // fix taken from https://github.com/OpenZeppelin/zeppelin-solidity/commit/d9b9ed227b175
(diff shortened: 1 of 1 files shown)


nberger:
Here's the commit in zeppelin removing these short address checks: https://github.com/OpenZeppelin/zeppelin-solidity/issues/261

I'm not even sure what's the way to protect from this kind of attack now, if anyone finds information please share it here. It supposedly has to be in upper layers (the client?) but I'm not 100% sure what that means.

PR #59, opened 2017-07-11
Adding contract interfaces

AugustoL:
Created contract interfaces, this help us and future developers to use the Lif contracts with only requesting the interface of the contracts, and also help us to use the interface between the crowdsale and future payments contracts.

### DIFF
--- a/contracts/LifInterface.sol
+++ b/contracts/LifInterface.sol
@@ -1,0 +1,15 @@
+pragma solidity ^0.4.11;
+
+import 'zeppelin-solidity/contracts/token/ERC20.sol';
+
+/**
+ * @title Lif token interface
+ */
+contract LifInterface is ERC20 {
+  uint public maxSupply;
+  function approveData(address spender, uint value, bytes data, bool doCall);
+  function transferData(address to, uint value, bytes data, bool doCall);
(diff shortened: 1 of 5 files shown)


contracts/LifInterface.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.4.11;
+
+import 'zeppelin-solidity/contracts/token/ERC20.sol';
+
+/**
+ * @title Lif token interface
+ */
+contract LifInterface is ERC20 {
+  uint public maxSupply;
```
nberger: what's the goal of having maxSupply in the interface?

AugustoL: To see the difference about the total tokens ins distributions and the maximun supply, in maxSupply we take in count the future payments for example, and the users would be able to see how much of the total tokens where distributed. They would be able to see how after the crowdsale during 4 years we reach the 100% token distribution between token buyers and founders.
I think its useful to have the maxSupply determined and written in the contract. So the traders, users, developers can access it easily. And also calculate the % distributed of tokens.

nberger: Makes sense, thanks

PR #60, opened 2017-07-16
[WIP] Propery-based tests

### LINKED ISSUE #25: Add stateful property-based test
Use a property-based testing library to generate sequences of transactions , run them and verify invariants/properties that must hold (for example a `submitBid` call should fail if the amount == 0, but it should go through and add the bid to the CrowdsaleStage, except if the token is stopped in which case it should also fail.

The main benefit of this kind of test suite is that the input data has some level of randomness, and also the sequence of commands is arbitrary so we end testing combinations of transactions that we wouldn't have time to write tests for (even if we came up with the same combinations "by hand")

nberger:
Adds a property based test which generates a crowdsale and a sequence of commands to run on that crowdsale. It runs the commands, checking some invariants/properties that must be true on every command that is run.

The interesting thing is that any sequence of commands can be generated, in ways that unit tests would not cover (because it's too expensive to write all the possible combinations so we only write a few specific combinations that we as developers think should cover most of the important/edge cases).

Example of a property:

> `submitBid` must throw an exception if the status of the crowdsale is not equal to 2 (which means that the crowdsale is started)

### Commands to implement:

- [x] waitBlock
- [x] checkPrice
- [x] submitBid
- [x] setStatus
- [ ] addPresalePayment
- [ ] distributeTokens
- [x] checkCrowdsale

### Verify resulting state:

- [x] Verify status (stopped, started, finished)
- [ ] Verify token ether balance
- [ ] Verify token tokens balance
- [ ] Verify contract ether balance
- [ ] Verify contract tokens balance
- [ ] Verify each account tokens balance
- [ ] Verify each account ether balance (add claimEth command?)
- [ ] Verify founders future payments
- [ ] Verify presale payments

Fixes #25

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,4 @@ cache:
 script:
   - set -e
   - jshint --show-non-errors test/*.js || echo "Errors in jshint ignored bc it doesn't support async/await"
-  - yarn test
+  - GEN_TESTS_QTY=20 yarn test
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -213,21 +213,22 @@ contract LifCrowdsale is Ownable, PullPayment {
       if (tokensSold.add(presaleTokens).add(tokensQty) > totalTokens)
         throw;
 
-      if (weiRaised.add(weiCost) <= maxCap) {
+      if (weiRaised.add(weiCost) > maxCap)
+        throw;
 
-        if (weiChange > 0)
-          safeSend(msg.sender, weiChange);
+      //
+      // bid is accepted from here
+      //
 
-        lastPrice = tokenPrice;
-        weiPayed[msg.sender] = weiCost;
-        tokens[msg.sender] = tokensQty;
-        weiRaised = weiRaised.add(weiCost);
-        tokensSold = tokensSold.add(tokensQty);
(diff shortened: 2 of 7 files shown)


nberger:
Last commit (f383c34) fixes a bug found by the test. Output was:

> Failed after 11 tests and 1 shrinks. rngState: 854df87bac5169302e; Counterexample: {commands: [{"type":"waitBlock"}, {"type":"waitBlock"}, {"type":"waitBlock"}, {"type":"setStatus","status":3}, {"type":"checkPrice"}], crowdsale: {startPriceEth: 16, changePerBlock: 37, changePriceEth: 45, minCapEth: 23, maxCapEth: 32, maxTokens: 40, presaleBonusRate: 23, ownerPercentage: 27}}; Exception: expected price is different! Expected: 16000000000000000000, actual: 0. blocks: 98, start/end: 93/98: expected 16000000000000000000 to equal 0

So basically the last block (`endBlock`) was not being considered part of the crowdsale, but it should IMO

nberger:
@AugustoL: I think we are at a point where it would be of value to get this tests in master, even though not all transactions are included yet. Could you please take a look so if it looks good we could merge it and keep adding the other transactions on further PRs?

PR #63, opened 2017-07-24
Remove ropsten from truffle.js from now

nberger:
We'll add it later, but for now it just it's trouble when trying
to run tests and network is offline


kvbik:
+1

PR #69, opened 2017-07-29
Add TokensIssued event

nberger:
Can help in debugging but also might have other use cases in the future


contracts/LifToken.sol
```diff
@@ -158,6 +161,8 @@ contract LifToken is LifInterface, LifDAOInterface, Ownable, PullPayment {
       allowed[address(this)][owner] = allowed[address(this)][owner].add(formatedBalance);
```
nberger: Sounds good, let's create a separate issue for that

PR #79, opened 2017-08-09
Add new token and crowdsale. Crowdsale is fixed price and uncapped now

AugustoL:


contracts/LifToken.sol
```diff
@@ -1,472 +1,17 @@
...
+pragma solidity ^0.4.13;
 
 import "zeppelin-solidity/contracts/ownership/Ownable.sol";
-import "zeppelin-solidity/contracts/payment/PullPayment.sol";
-import 'zeppelin-solidity/contracts/math/SafeMath.sol';
-import './LifInterface.sol';
-import './LifDAOInterface.sol';
+import "zeppelin-solidity/contracts/token/MintableToken.sol";
+import "zeppelin-solidity/contracts/math/SafeMath.sol";
+import "zeppelin-solidity/contracts/lifecycle/Pausable.sol";
```
kvbik: do we have to import the `SafeMath` and `Ownable` too? i don't see it used in the code.

AugustoL: seems that we dont have to.

kvbik:
good job on friday!! what do we miss here to merge it to `master`?:)))

nberger:
@kvbik I think not much! Things missing, from the top of my head:

- add presale payments
- transfer funds to foundation wallet as soon as minCap is reached and until market maker low cap is hit
- all the market maker stuff
- add gen test command for pausing the token (pausing crowdsale has a command already)
- add more state assertions in gen test
- add a gen test command for the crowdsale finalize op

But we can work on those in further PRs so yeah next step is to cleanup the commits here and merge :)

kvbik:
100% agree, looking forward!

nberger:
Build looks like if it's broken here, but it's not. Last build was https://travis-ci.org/windingtree/LifToken/builds/264279082 which is green, so I guess it's just that travis and github are not friends today :)

PR #80, opened 2017-08-10
Some tweaks/refactors in prop based test

nberger:


kvbik:
why is this one not merged?

nberger:
@kvbik because build is broken

nberger:
We can reject this one as these changes where included in #79 :)

PR #86, opened 2017-08-14
Finalize crowdsale command

AugustoL:
Added finalize crowdsale command and crowdsaleEnded variable in state.

### DIFF
--- a/test/CrowdsaleGenTest.js
+++ b/test/CrowdsaleGenTest.js
@@ -105,6 +105,7 @@ contract('LifCrowdsale Property-based test', function(accounts) {
     let shouldThrow = (nextBlock < startBlock) ||
       (nextBlock > endBlock2) ||
       (state.crowdsalePaused) ||
+      (state.crowdsaleFinalized) ||
       (weiCost == 0);
 
     try {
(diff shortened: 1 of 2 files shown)


nberger:
Seems like the first commit should be thrown away

test/CrowdsaleGenTest.js
```diff
@@ -154,12 +158,30 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
     }
 
     return state;
   };
 
+  let runFinalizeCrowdsaleCommand = async (command, state) => {
+    let shouldThrow = (state.crowdsalePaused == command.pause) ||
+      (command.fromAccount != 0) ||
+      !state.crowdsaleContract.isFinalized ||
+      !(await state.crowdsaleContract.hasEnded());
```
nberger: we should do this kind of checks on the on-memory state, not on the contract. So if `hasEnded` is not in the on-memory state yet, we should add it

test/CrowdsaleGenTest.js
```diff
@@ -154,12 +158,30 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
       if (!shouldThrow)
-        throw(new ExceptionRunningCommand(e, state, command));   
+        throw(new ExceptionRunningCommand(e, state, command));
     }
 
     return state;
   };
 
+  let runFinalizeCrowdsaleCommand = async (command, state) => {
+    let shouldThrow = (state.crowdsalePaused == command.pause) ||
```
nberger: this command doesn't have a `pause`, please check this condition

test/CrowdsaleGenTest.js
nberger: the same applies to isFinalized: it should be checked against the in-memory state, not on the crowdsale contract.

test/CrowdsaleGenTest.js
```diff
@@ -154,12 +159,28 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
       if (!shouldThrow)
-        throw(new ExceptionRunningCommand(e, state, command));   
+        throw(new ExceptionRunningCommand(e, state, command));
     }
 
     return state;
   };
 
+  let runFinalizeCrowdsaleCommand = async (command, state) => {
+    let shouldThrow = state.crowdsaleFinalized || state.crowdsalePaused ||
```
nberger: can you please put every condition in a new line? It's easier to read I think

nberger: I think it should also throw if `command.fromAccount != 0`, no?

nberger: Oh, you are right, anyone can finalize it, so that condition is not needed

nberger:
LGTM :+1:
Feel free to squash everything to a single commit, in this case it seems that should be enough

PR #89, opened 2017-08-14
Add pause token command

### LINKED ISSUE #83: Add prop-based test command for pausing the token
We have to think about the logic on pausing the token a bit more though. We are already pausing the token to prevent transfers while the Crowdsale is ongoing. However, we have to allow minting during the Crowdsale, but at the same time there should be a way to pause the minting too.

nberger:
Fixes #83

### DIFF
--- a/test/CrowdsaleGenTest.js
+++ b/test/CrowdsaleGenTest.js
@@ -51,6 +51,11 @@ contract('LifCrowdsale Property-based test', function(accounts) {
     pause: jsc.bool,
     fromAccount: accountGen
   });
+  let pauseTokenCommandGen = jsc.record({
+    type: jsc.constant("pauseToken"),
+    pause: jsc.bool,
+    fromAccount: accountGen
+  });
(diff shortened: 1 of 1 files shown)


test/CrowdsaleGenTest.js
```diff
@@ -106,7 +111,7 @@ contract('LifCrowdsale Property-based test', function(accounts) {
       (nextBlock > endBlock2) ||
       (state.crowdsalePaused) ||
       (state.crowdsaleFinalized) ||
-      (weiCost == 0);
+      (command.eth == 0);
```
AugustoL: How is this change related to the pause token command?

nberger: not really very related :)
Do you think I should do a separate commit, or even a separate PR?

AugustoL: A separate commit will be enough :)

nberger: @AugustoL ok, better now? :)

PR #91, opened 2017-08-14
Add private presale tokens on LifCrowdsale with test command

AugustoL:
Add function for the owner of the crowdsale to add tokens to addresses, this tokens are taken in count as private presale payments. The owner has a maximun amount of tokens that he can add, thats why I also added a maxPrivatePresaleTokens variable on the contract.

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -137,15 +137,15 @@ contract LifCrowdsale is Ownable, Pausable {
     TokenPurchase(msg.sender, beneficiary, weiAmount, tokens);
   }
 
-  function addPrivatePresaleTokens(address beneficiary, uint256 totalWei) onlyOwner {
+  function addPrivatePresaleTokens(address beneficiary, uint256 weiSent) onlyOwner {
     require(block.number < startBlock);
     require(beneficiary != address(0));
-    require(totalWei > 0);
+    require(weiSent > 0);
 
-    uint256 tokens = totalWei.mul(privatePresaleRate);
+    uint256 tokens = weiSent.mul(privatePresaleRate);
 
     uint256 presaleWei = totalPresaleWei;
-    require(presaleWei.add(totalWei) <= maxPresaleWei);
+    require(presaleWei.add(weiSent) <= maxPresaleWei);
 
     totalPresaleWei.add(tokens);
 
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -144,10 +144,9 @@ contract LifCrowdsale is Ownable, Pausable {
 
     uint256 tokens = weiSent.mul(privatePresaleRate);
 
-    uint256 presaleWei = totalPresaleWei;
-    require(presaleWei.add(weiSent) <= maxPresaleWei);
+    require(totalPresaleWei.add(weiSent) <= maxPresaleWei);
 
-    totalPresaleWei.add(tokens);
+    totalPresaleWei.add(weiSent);
 
     token.mint(beneficiary, tokens);
   }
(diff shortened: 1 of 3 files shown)


contracts/LifCrowdsale.sol
```diff
@@ -126,10 +137,17 @@ contract LifCrowdsale is Ownable, Pausable {
     TokenPurchase(msg.sender, beneficiary, weiAmount, tokens);
   }
 
-  function addPresaleTokens(address beneficiary, uint256 tokens) onlyOwner {
+  function addPrivatePresaleTokens(address beneficiary, uint256 totalWei) onlyOwner {
     require(block.number < startBlock);
     require(beneficiary != address(0));
-    require(tokens > 0);
+    require(totalWei > 0);
```
nberger: I think we should change the naming here. `totalWei` in this function is the amount of wei that corresponds to this presale payment. But `maxPresaleWei` is the sum of the wei sent in all the presale payments. The naming doesn't help in understanding the relation between those two values. I think we should rename `totalWei` to `weiSent` and `totalPresaleWei` to `totalPresaleWeiReceived`

What do you think @AugustoL?

contracts/LifCrowdsale.sol
```diff
@@ -126,10 +137,17 @@ contract LifCrowdsale is Ownable, Pausable {
...
+
+    uint256 tokens = totalWei.mul(privatePresaleRate);
+
+    uint256 presaleWei = totalPresaleWei;
```
nberger: why a separate `presaleWei` variable if we can just use `totalPresaleWei` (or whatever new name it gets)?

AugustoL: Because I need to calculate the totalWei that will be adding the weiSent, if the totalAmount isnt higher than the maxPresaleWei I can add the weiSent to the totalPresaleWei and edit that variable,

contracts/LifCrowdsale.sol
AugustoL: I think renaming `totalWei` to `weiSent` would be good, but for `totalPresaleWei` there is no need to rename it, that is the total wei raised by presale, and it cant be higher than `maxPresaleWei`

test/CrowdsaleGenTest.js
```diff
@@ -208,46 +209,35 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
-          throw(e);
-      }
-    }
+  let runAddPrivatePresalePaymentCommand = async (command, state) => {
+
+    let crowdsale = state.crowdsaleData,
+      { startBlock, endBlock2, maxPresaleWei, privatePresaleRate } = crowdsale,
+      nextBlock = web3.eth.blockNumber + 1,
+      tokens = command.tokens,
+      totalWei = web3.toWei( tokens * ( 1 / privatePresaleRate)),
```
nberger: should we rename it to `presaleWei` or `weiToSend` or something like that? :)

AugustoL: yep, doing it

contracts/LifCrowdsale.sol
```diff
@@ -30,6 +33,10 @@ contract LifCrowdsale is Ownable, Pausable {
 
   uint256 public tokensSold;
 
+  uint256 public totalPresaleWei;
```
nberger: let's put a comment describing what this var means, and the same for the next var? This could be smth like "total amount of wei received as presale payments (both private and public)"

test/CrowdsaleGenTest.js
```diff
@@ -278,6 +269,9 @@ contract('LifCrowdsale Property-based test', function(accounts) {
     */
     help.debug("checking purchases total wei, purchases:", JSON.stringify(state.purchases));
     assert.equal(_.sumBy(state.purchases, (b) => b.wei), parseInt(await crowdsale.weiRaised.call()));
+
+    // Check presale tokens sold
+    assert.equal(state.totalPresaleWei, parseInt(await crowdsale.totalPresaleWei.call()));
```
nberger: nice :)

PR #92, opened 2017-08-15
Update LifToken

AugustoL:
* Add transferData, approveData and transferDataFrom methods on token.
* Added unit tests for LifToken.
* Removed unnecessary dependency used in old units tests for LifToken.

### DIFF
--- a/contracts/LifToken.sol
+++ b/contracts/LifToken.sol
@@ -12,4 +12,54 @@ contract LifToken is MintableToken, Pausable {
 
   // Token decimals
   uint public constant DECIMALS = 18;
+
+  // Edit of the ERC20 token events to support data argument
+  event TransferData(address indexed from, address indexed to, uint value, bytes data);
+  event ApprovalData(address indexed from, address indexed spender, uint value, bytes data);
+
+  //ERC20 token approve method with data call.
+  function approveData(address spender, uint value, bytes data) {
+
+    require(spender != address(this));
+
+    allowed[tx.origin][spender] = value;
+
+    if (spender.call(data))
+      ApprovalData(tx.origin, spender, value, data);
+
+  }
+
+  // ERC20 transfer method with data call.
+  function transferData(address to, uint value, bytes data) {
+
+    require(to != address(this));
+
+    // If transfer have value process it
+    if (value > 0) {
+      balances[tx.origin] = balances[tx.origin].sub(value);
+      balances[to] = balances[to].add(value);
+    }
+
+    if (to.call(data))
+      TransferData(tx.origin, to, value, data);
+
+  }
+
+  // ERC20 transferFrom method with data call.
+  function transferDataFrom(address from, address to, uint value, bytes data) {
+
+    require(to != address(this));
+
+    // If transfer have value process it
+    if (value > 0) {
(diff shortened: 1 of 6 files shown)


contracts/LifToken.sol
```diff
@@ -12,4 +12,54 @@ contract LifToken is MintableToken, Pausable {
 
   // Token decimals
   uint public constant DECIMALS = 18;
+
+  // Edit of the ERC20 token events to support data argument
```
nberger: Instead of "Edit of the ERC20 token events...", I think it should be something like "Addition on top of ERC20 token events..." or "Variation of ERC20 token events ..."

AugustoL: "Extra events based in ERC20 events"
and for functions:
"Extra methods transferData based on ERC20 transfer method, transferData allow the user to transfer value+data on transaction"
what do you think @nberger ?

nberger: I'd do this:

"transferData is an addition to ERC20 token methods. It allows to transfer value and data on each transaction"

contracts/LifToken.sol
```diff
@@ -12,4 +12,54 @@ contract LifToken is MintableToken, Pausable {
...
+
+    allowed[tx.origin][spender] = value;
+
+    if (spender.call(data))
+      ApprovalData(tx.origin, spender, value, data);
+
+  }
+
+  // ERC20 transfer method with data call.
+  function transferData(address to, uint value, bytes data) {
```
nberger: ERC20 `transfer` returns bool, shouldn't we do the same here?

test/helpers.js
```diff
@@ -67,126 +67,66 @@ module.exports = {
...
-      status: parseInt(proposal[4]),
-      creationBlock: parseInt(proposal[5]),
-      maxBlock: parseInt(proposal[6]),
-      agePerBlock: parseInt(proposal[7]),
-      votesNeeded: parseInt(proposal[8]),
-      actionData: proposal[9],
-      totalVotes: parseInt(proposal[10])
-    };
-    console.log('['+parsedProposal.id+'] To: '+parsedProposal.target+', Value: '+parsedProposal.value +', MaxBlock: '+parsedProposal.maxBlock+', Desc: '+parsedProposal.description+', Status: '+parsedProposal.status, ', Votes: ',parsedProposal.totalVotes,'/',parsedProposal.votesNeeded);
+  simulateCrowdsale: async function(rate, balances, accounts) {
```
nberger: why don't we keep this function in test/LifToken.js? We are only using it from there for now. If we need it somewhere else, we'll see :). We might decide to copy/paste it, or to extract smaller parts of it. For me helpers should be for general and more simple stuff, this could go to something like `factories.js` but until then I feel better to have it in the same file (easier to work with it).

Also I'm wondering if we shouldn't try and build a property based test for LifToken...

what do you think @AugustoL?

AugustoL: I agree on both things, lets have the simulateCrowdsale on the LifToken.js file, and yes, we will need property based tests for LifToken too :)
For now I will just move the simulateCrowdsale, then we can do another PR for the prop based tests on LifToken.

Issue #81, opened 2017-08-14
Add public presale on LifCrowdsale contract and tests for it.


nberger:

AugustoL:
We have to add the necessary functions and variables on the LifCrowdsale contract to run a public presale of a maximum amount of tokens before the ICO.

I think that we would need:
* startPresaleBlock & endPresaleBlock.
* maxTokensPresale.
* publicPresaleTokensSold.
* buyPresaleTokens(address beneficiary)
* modify the fallback function to execute the buyPresaleTokens it its between the right blocks.

nberger:
Done in #94

PR #94, opened 2017-08-16
Add Public Presale

AugustoL:


contracts/LifCrowdsale.sol
```diff
@@ -11,6 +11,10 @@ contract LifCrowdsale is Ownable, Pausable {
   // The token being sold
   LifToken public token;
 
+  // start and end of the public presale
+  uint256 public startPublicPresaleBlock;
+  uint256 public endPublicPresaleBlock;
```
nberger: these should be `publicPresaleStartBlock` and `publicPresaleEndBlock`, the same for any other variables: `publicPresaleWhatever`, `privatePresaleWhatever`, etc

contracts/LifCrowdsale.sol
```diff
@@ -23,6 +27,8 @@ contract LifCrowdsale is Ownable, Pausable {
   // how much wei a token unit costs to a buyer, during the private presale stage
   uint256 public privatePresaleRate;
 
+  // how much wei a token unit costs to a buyer, during the public presale
+  uint256 public ratePublicPresale;
```
nberger: please rename to `publicPresaleRate`

contracts/LifCrowdsale.sol
AugustoL: @nberger for what other variables? for private presale we dont need to store block range, the private presale can happen only between contract creation and before `publicPresaleStartBlock`

nberger: actually maybe we should add a privatePresaleEndBlock... and other variables are whatever other variables the publicPresale uses, for example `publicPresaleRate`

AugustoL: we have a privatePresaleRate, thats the only thing we need for the private presale, we dont need more. And thats for private, for public here should I make any change?

nberger: I think we should rename these variables as I said above, they should be:

`publicPresaleStartBlock`
`publicPresaleEndBlock`

Also, I think we might need a `privatePresaleEndBlock`, to see the finish date of the private presale. Or does it have to be exactly the same as `publicPresaleStartBlock`?

AugustoL: I will rename them, for me there is no need to have a endBlock for private presale, whats wrong in using the public presale start block?

nberger: Yeah, it's probably ok to just use the public presale start block :+1:

contracts/LifCrowdsale.sol
```diff
@@ -131,17 +150,23 @@ contract LifCrowdsale is Ownable, Pausable {
     // calculate token amount to be created
     uint256 tokens = weiAmount.mul(rate);
 
-    // update state
-    weiRaised = weiRaised.add(weiAmount);
-    purchases[beneficiary] = weiAmount;
-    tokensSold = tokensSold.add(tokens);
+    // check if we are in the public presale block range
+    if (rate == publicPresaleRate) {
```
nberger: I think we should have separate functions for private presale, public presale, and ICO tokens purchases. Maybe we can keep a single getRate fn though... but the stages should be explicit when purchasing

AugustoL: I think its better this way because they can use the fallback function at any time during the public presale and ICO, will be easier for investors to send the funds. Also more secure, if we have to publish instructions about how to invest on the public presale building a function will be more difficult for the investor and he can make a mistake. Using fallback for me is the way to go. If we decide to change it we can do it on another PR.

nberger: we can still make the fallback function call one or another, to make it easier for people who want to use that way

AugustoL: You are right, so we can check if its in public presale or ICO on the fallback function and call buyTokens or buyTokensPresale, what do you think?

nberger: yes, that sounds ok to me :+1:

AugustoL: :+1:

kvbik:
closes #81

contracts/LifCrowdsale.sol
```diff
@@ -162,11 +206,16 @@ contract LifCrowdsale is Ownable, Pausable {
   }
 
   // @return true if the transaction can buy tokens
-  function validPurchase() internal constant returns (bool) {
+  function validPurchase(bool fromPresaleBuy) internal constant returns (bool) {
```
nberger: My last ask is to separate this function into 2 separate fns: `validPurchase` and `validPresalePurchase`, they will be more focused in one task

AugustoL: :+1: your wish would be granted

nberger:
:shipit:

PR #99, opened 2017-08-19
Market maker getCurrentPeriodIndex() implemented with tests

AugustoL:


contracts/LifMarketMaker.sol
```diff
@@ -59,10 +59,11 @@ contract LifMarketMaker is Ownable {
   }
 
   function calculateDistributionPeriods(
-    uint256 startBlock, uint8 totalPeriods, uint256 blocksPerPeriod
+    uint256 _startBlock, uint8 totalPeriods, uint256 blocksPerPeriod
```
nberger: if we use `_` to name parameters, let's do it for all of them. But also, as this function already uses (mutates) the `distributionPeriods` variable, feel free to not pass any parameters and just use the already defined MarketMaker state variables :)

PR #100, opened 2017-08-19
Add claimEth and getMaxClaimableEthAmount methods

nberger:


contracts/LifMarketMaker.sol
```diff
@@ -133,20 +140,22 @@ contract LifMarketMaker is Ownable {
...
 
-  // Get the total amount of wei tat the foundation can claim in the current distribution period
-
-  // TODO Calculate the total amount of wei that the foundation can withdraw, not only of current period
-  function getFoundationWei() constant public returns (uint256) {
+  // Get the maximum amount of wei that the foundation can claim, without discounting what it
+  // claimed already (so the actual amount that it can claim can be lower). It's a portion of
+  // the ETH that was not claimed by token holders plus the profits made by the market maker
+  // by buying and selling tokens
+  function getMaxClaimableEthAmount() constant public returns (uint256) {
```
AugustoL: This function returns the unit in Wei, should be called `getMaxClaimableWeiAmount` ?

nberger: Yes! good catch, will fix now

nberger: fixed in dc7beac

contracts/LifMarketMaker.sol
```diff
@@ -33,11 +33,15 @@ contract LifMarketMaker is Ownable {
   // The price at which the market maker buys tokens at the beginning of its lifetime
   uint256 public initialBuyPrice = 0;
 
+  uint256 percentagesFactor = 10000;
```
AugustoL: Should be called `PERCENTAGE_FACTOR` and be a public constant?

PR #103, opened 2017-08-19
Market maker update all unit tests

kvbik:
json objects instead of comments and couple helper functions


kvbik:
need to make the javascript work, but feel free to check the numbers are ok

test/MarketMaker.js
```diff
@@ -145,146 +145,168 @@ contract('marketMaker', function(accounts) {
...
+    // MMETH = 38,    TP = 68,   MMT = 100, TC = 0,   SP = 2163 mETH/Lif, BP = 800,  CL 60%, maxClaimable = 0,   claimed = 30
+    checkScenarioProperties({
+      marketMakerEthBalance: 38, marketMakerLifBalance: 100, totalProfit: 68,
+      customerEthBalance: 632, customerLifBalance: 0,
+      marketMakerSellPrice: 2163, marketMakerBuyPrice: 800,
+      claimablePercentage: 60, maxClaimableEth: 0, totalClaimedEth: 30
+    });
+
+    // Buy 100 tokens
+    await token.approve(customer, 100);
```
nberger: no need to approve when buying tokens from the market maker

test/MarketMaker.js
```diff
@@ -145,146 +145,168 @@ contract('marketMaker', function(accounts) {
...
-  // Claimable month: 366
-  // Claimable total: 19+154+366 = 543
-  // Claim(243);
-  // Claimable total: 300
-  // claimableUpdatedMonth: 3
-
-  // Month 4:
-  // Claimable month: 651
-  // Claimable total: 300+651 = 951
+  var checkScenarioProperties = async function(data, mm, customer) {
```
AugustoL: periodIndex is missing here

kvbik: there is a helper function `waitForMonth` called before in the test

AugustoL: yes, but we want to check that we are in the correct period after we wait, right?

kvbik: that means we do not believe the helper? and we are checking it all over again. not sure...

PR #108, opened 2017-08-19
Market maker sellPrice calculation implemented with unit tests

AugustoL:


contracts/LifMarketMaker.sol
```diff
@@ -63,21 +68,25 @@ contract LifMarketMaker is Ownable {
     totalPeriods = _totalPeriods;
     foundationAddr = _foundationAddr;
     initialWei = msg.value;
-    initialBuyPrice = initialWei.div(lifToken.totalSupply());
+    initialBuyPrice = initialWei
+      .div(lifToken.totalSupply().div(PRICE_FACTOR))
```
nberger: we should first `mul(PRICE_FACTOR)` and then `div(lifToken.totalSupply....)` to keep more precision

contracts/LifMarketMaker.sol
```diff
@@ -63,21 +68,25 @@ contract LifMarketMaker is Ownable {
...
+      .mul(PRICE_FACTOR);
+    initialSellPrice = initialBuyPrice.div(PRICE_FACTOR).mul(initialPriceSpread);
```
nberger: Why do we `div(PRICE_FACTOR)`? Is it because `initialPriceSpread` is already multipled by it? If that's true, I think we should first `mul`, then `div`

contracts/LifMarketMaker.sol
```diff
@@ -97,22 +106,64 @@ contract LifMarketMaker is Ownable {
...
 
     uint256 accumDistribution = 0;
-    uint256 periodStartBlock = startBlock;
+    uint256 deltaDistribution = 0;
+    uint256 startBlockPeriod = startBlock;
 
     for (uint8 i = 0; i < totalPeriods; i++) {
-      uint256 deltaDistribution;
+
+      require(marketMakerPeriods.length <= i);
```
nberger: what are we validating here? seems like we can just `require(marketMakerPeriods.length == 0)` before the for loop

contracts/LifMarketMaker.sol
AugustoL: initiatlSpread its already mulltiplied by PRICE_FACTOR, and yes, I will change the order.

contracts/LifMarketMaker.sol
```diff
@@ -122,21 +173,21 @@ contract LifMarketMaker is Ownable {
...
-  function getSellRate() public constant returns (uint256) {
+  function getSellPrice() public constant returns (uint256) {
 
-    uint256 foundationWei = 0; //getFoundationWei();
+    uint256 periodIndex = getCurrentPeriodIndex();
 
-    uint256 currentMarketMakerWei = this.balance.sub(foundationWei);
+    uint256 sellPriceIncrement = initialSellPrice
+      .div(PRICE_FACTOR)
+      .mul(PRICE_FACTOR.add(marketMakerPeriods[periodIndex].accumSellPriceIncrement));
```
nberger: first mul, then div

contracts/LifMarketMaker.sol
AugustoL: Yes, the validation is to call this function only once, we can do it before the loop. I will do that.

contracts/LifMarketMaker.sol
AugustoL: :+1:

contracts/LifMarketMaker.sol
```diff
@@ -162,7 +213,7 @@ contract LifMarketMaker is Ownable {
   }
 
   function() payable {
-    buyLif();
+    getTokens();
   }
 
   function getTokens() payable {
```
nberger: In this function, `getBuyPrice` is being used when `getSellPrice` should be used instead. Also we have to add profit calculation here, but if you want to leave it out from this PR that's fine to me

test/MarketMaker.js
```diff
@@ -162,6 +215,68 @@ contract('marketMaker', function(accounts) {
     await help.waitToBlock(startBlock+blocksPerPeriod*numberOfMonth, accounts);
   };
 
+  it("Should return the correct sellPrice on every period", async function() {
```
nberger: there are no assertions in this test, they are commented out

contracts/LifMarketMaker.sol
AugustoL: Changed function name on fallback, and on for getSellPrice on getTokens(), I leave the profit for another PR.

test/MarketMaker.js
AugustoL: Assertions added and test working :)

contracts/LifMarketMaker.sol
```diff
@@ -63,21 +68,28 @@ contract LifMarketMaker is Ownable {
     totalPeriods = _totalPeriods;
     foundationAddr = _foundationAddr;
     initialWei = msg.value;
-    initialBuyPrice = initialWei.div(lifToken.totalSupply());
+    initialBuyPrice = initialWei
+      .mul(PRICE_FACTOR)
+      .div(lifToken.totalSupply().div(PRICE_FACTOR));
```
nberger: I think we should remove the `div(PRICE_FACTOR)`, just keep the `mul` in previous line

contracts/LifMarketMaker.sol
```diff
@@ -97,22 +109,64 @@ contract LifMarketMaker is Ownable {
...
       } else {
         deltaDistribution = deltas48[i];
       }
+
       accumDistribution = accumDistribution.add(deltaDistribution);
-      uint256 periodEndBlock = periodStartBlock.add(blocksPerPeriod).sub(1);
 
-      distributionPeriods.push(DistributionPeriod(
-        periodStartBlock, periodEndBlock, deltaDistribution, accumDistribution
+      uint256 endBlockPeriod = startBlockPeriod.add(blocksPerPeriod).sub(1);
```
nberger: as we moved the `deltaDistribution` var declaration out from the for loop, we should also move `endBlockPeriod` out for consistency, no?

contracts/LifMarketMaker.sol
```diff
@@ -97,22 +109,64 @@ contract LifMarketMaker is Ownable {
...
+      17257, 18430, 19614, 20810, 22019,
+      23239, 24471, 25716, 26973, 28243,
+      29525, 30820, 32129, 33450, 34784,
+      36132, 37494, 38869, 40257, 41660,
+      43076, 44507, 45952, 47412, 48886,
+      50375, 51878, 53397, 54931, 56481,
+      58045, 59626
+    ];
+
+    uint256 accumSellPriceIncrement = 0;
```
nberger: this var is not used I think

contracts/LifMarketMaker.sol
```diff
@@ -122,21 +176,21 @@ contract LifMarketMaker is Ownable {
...
+    uint256 periodIndex = getCurrentPeriodIndex();
 
-    uint256 currentMarketMakerWei = this.balance.sub(foundationWei);
+    uint256 sellPriceIncrement = initialSellPrice
+      .mul(PRICE_FACTOR.add(marketMakerPeriods[periodIndex].accumSellPriceIncrement))
+      .div(PRICE_FACTOR);
 
-    uint256 sellRate = lifToken.totalSupply()
-      .div(currentMarketMakerWei);
+    return sellPriceIncrement;
```
nberger: The name of the variable is misleanding. `sellPriceIncrement` seems to be about some increment, but this is about the actual sellPrice. I'd just remove the var and return the value right away, that solves the naming issue :)

test/MarketMaker.js
```diff
@@ -162,6 +215,69 @@ contract('marketMaker', function(accounts) {
...
+
+      // help.debug(
+      //   parsePrice(parseFloat(initialSellPrice*(1.01**i)))
+      //   ,' == ',
+      //   parsePrice(parseFloat(await mm.getSellPrice()))
+      // );
+      assert.approximately(
+        parsePrice(parseFloat(initialSellPrice*(1.01**i))),
+        parsePrice(parseFloat(await mm.getSellPrice())),
+        1, 'wrong sell price in contract on period index'+ i
```
nberger: we are using a delta of 1. That's the precision we have? Can't we use a lower delta? What are the values typically being compared here? I was thinking a price could be for example 7.82352. Delta of 1 means any value in the range `6.82352 < x < 8.82352` would be ok for this assertion, that's a too low precision

AugustoL: The difference allowed here its not on the delta, its on the sellPrice received and the calculated outside the contract, and over the price in wei, 1 wei of difference between prices its a very high precision. Wen I format the price to Lif outside the contract the price its the same on the first 4 decimals.

```
Sell Price on period 0 :in memory 8.04000000000000032e-14  == on contract 8.04000000000000032e-14
Sell Price on period 1 :in memory 8.12039999999999966e-14  == on contract 8.12039999999999966e-14
Sell Price on period 2 :in memory 8.20160000000000011e-14  == on contract 8.20160000000000011e-14
Sell Price on period 3 :in memory 8.28360000000000042e-14  == on contract 8.28360000000000042e-14
Sell Price on period 4 :in memory 8.36640000000000058e-14  == on contract 8.36640000000000058e-14
Sell Price on period 5 :in memory 8.45010000000000026e-14  == on contract 8.45010000000000026e-14
Sell Price on period 6 :in memory 8.53459999999999980e-14  == on contract 8.53459999999999980e-14
Sell Price on period 7 :in memory 8.61990000000000046e-14  == on contract 8.61990000000000046e-14
Sell Price on period 8 :in memory 8.70609999999999937e-14  == on contract 8.70609999999999937e-14
Sell Price on period 9 :in memory 8.79320000000000034e-14  == on contract 8.79309999999999941e-14
Sell Price on period 10 :in memory 8.88109999999999989e-14  == on contract 8.88109999999999989e-14
Sell Price on period 11 :in memory 8.96990000000000023e-14  == on contract 8.96990000000000023e-14
Sell Price on period 12 :in memory 9.05960000000000009e-14  == on contract 9.05960000000000009e-14
Sell Price on period 13 :in memory 9.15019999999999947e-14  == on contract 9.15019999999999947e-14
Sell Price on period 14 :in memory 9.24169999999999964e-14  == on contract 9.24169999999999964e-14
Sell Price on period 15 :in memory 9.33410000000000059e-14  == on contract 9.33410000000000059e-14
Sell Price on period 16 :in memory 9.42749999999999946e-14  == on contract 9.42739999999999980e-14
Sell Price on period 17 :in memory 9.52180000000000038e-14  == on contract 9.52169999999999945e-14
Sell Price on period 18 :in memory 9.61699999999999956e-14  == on contract 9.61689999999999990e-14
Sell Price on period 19 :in memory 9.71309999999999953e-14  == on contract 9.71309999999999953e-14
Sell Price on period 20 :in memory 9.81029999999999961e-14  == on contract 9.81029999999999961e-14
Sell Price on period 21 :in memory 9.90840000000000047e-14  == on contract 9.90840000000000047e-14
Sell Price on period 22 :in memory 1.00075000000000005e-13  == on contract 1.00073999999999996e-13
Sell Price on period 23 :in memory 1.01075000000000001e-13  == on contract 1.01075000000000001e-13
```

contracts/LifMarketMaker.sol
AugustoL: On the buyPrice? If it is on the buyPrice its unrelated to this PR.

nberger: How can it be unrelated to this PR since we are talking about a change introduced in this PR

AugustoL: My bad,

test/MarketMaker.js
```diff
@@ -162,6 +215,69 @@ contract('marketMaker', function(accounts) {
...
+    await mm.calculateDistributionPeriods({from: accounts[4]});
+    await mm.calculateSellPricePeriods({from: accounts[3]});
+
+    help.debug('MM balance:', parseInt( web3.eth.getBalance(mm.address) ));
+    help.debug('Start block', parseInt( await mm.startBlock.call() ));
+    help.debug('Blocks per period', parseInt( await mm.blocksPerPeriod.call() ));
+    help.debug('Foundation address', await mm.foundationAddr.call() );
+    help.debug('Initial Wei', parseInt( await mm.initialWei.call() ));
+    help.debug('Initial Buy Price', parseInt( await mm.initialBuyPrice.call() ));
+    help.debug('Initial Sell Price', parseInt( await mm.initialSellPrice.call() ));
```
nberger: can we put an assert on the expected `initialSellPrice`? It would be nice to be able to "see" that price here, without having to run the test to see it in the console

test/MarketMaker.js
```diff
@@ -162,6 +215,69 @@ contract('marketMaker', function(accounts) {
...
+    const blocksPerPeriod = 5;
+    const initialPriceSpread = 100500;
+
+    mm = await LifMarketMaker.new(
+      token.address, startBlock, blocksPerPeriod, 24,
+      accounts[1], initialPriceSpread,
+      {value: web3.toWei(8, 'ether'), from: accounts[0]}
+    );
+
+    const initialSellPrice = parseInt(await mm.initialSellPrice());
```
nberger: we should not use a number extracted from the contract as the basis for our expected results. I mean, in some scenarios that's valid, but not here, because this is exactly what we are testing: the price

test/MarketMaker.js
```diff
@@ -162,6 +215,70 @@ contract('marketMaker', function(accounts) {
...
+
+    await help.waitToBlock(startBlock+1);
+    for (var i = 0; i < 24; i++) {
+
+      help.debug('Sell price on period', i,
+        parseFloat(await mm.getSellPrice()/priceFactor)
+      );
+
+      assert.approximately(
+        parseFloat(initialSellPrice*(1.01**i))*priceFactor/priceFactor,
```
nberger: What's the reason behind `*priceFactor/priceFactor`? Can't we remove that term?

Perhaps we should keep all the numbers multiplied by priceFactor, so we don't have to divide what we get back from the contract by `priceFactor` like is done in the next line

Also, I'd still like to see some assertions (at least 1, maybe 3 would be nice) with concrete numbers (say what's the price at period 1, what's the price at period 14, what's the price at period 23)

test/MarketMaker.js
```diff
@@ -162,6 +215,70 @@ contract('marketMaker', function(accounts) {
...
+      23239, 24471, 25716, 26973, 28243,
+      29525, 30820, 32129, 33450, 34784,
+      36132, 37494, 38869, 40257, 41660,
+      43076, 44507, 45952, 47412, 48886,
+      50375, 51878, 53397, 54931, 56481,
+      58045, 59626
+    ];
+
+    await help.waitToBlock(startBlock+1);
+    for (var i = 0; i < 24; i++) {
```
nberger: why are we checking just 24 numbers if we have 48? Shouldn't we check all of them? Or, if it's because the MM was created with 24 periods, please remove the unused terms from accumIncrementPrice. By the way, do you use `accumIncrementPrice` at all?

AugustoL: You are right, going to change it to 48 months and removed unused `accumIncrementPrice`

test/MarketMaker.js
AugustoL: :+1:

PR #109, opened 2017-08-20
Make market maker test to work

nberger:
Let's take the draft test scenario we wrote all together and make it work doing real calls to `sendTokens`, `getTokens`, `claimEth` and all the expected assertions

/cc @kvbik @AugustoL

### DIFF
--- a/test/MarketMaker.js
+++ b/test/MarketMaker.js
@@ -1,10 +1,18 @@
 var help = require("./helpers");
 
+var BigNumber = web3.BigNumber;
+
+const should = require('chai')
+  .use(require('chai-bignumber')(BigNumber))
+  .should();
+
 var LifMarketMaker = artifacts.require("./LifMarketMaker.sol");
 var LifToken = artifacts.require("./LifToken.sol");
 var LifCrowdsale = artifacts.require("./LifCrowdsale.sol");
 
 const LOG_EVENTS = true;
+
+const gasPrice = new BigNumber(100000000000);
 
 contract('marketMaker', function(accounts) {
 
@@ -173,23 +181,6 @@ contract('marketMaker', function(accounts) {
     assert.equal(4, parseInt(await mm.getCurrentPeriodIndex()) );
   });
 
-  var checkScenarioProperties = async function(data, mm, customer) {
-    assert.equal(data.totalProfit, mm.totalProfit());
-    assert.equal(data.marketMakerEthBalance, web3.eth.getBalance(mm.address));
-    assert.equal(data.marketMakerLifBalance, lifToken.balanceOf(mm.address));
-    assert.equal(data.customerEthBalance, web3.eth.getBalance(customer));
-    assert.equal(data.customerLifBalance, lifToken.balanceOf(customer));
-    assert.equal(data.marketMakerSellPrice, mm.getSellPrice());
-    assert.equal(data.marketMakerBuyPrice, mm.getBuyPrice());
-    assert.equal(data.claimablePercentage, mm.getAccumulatedDistributedPercentage());
-    assert.equal(data.maxClaimableEth, mm.getMaxClaimableWei());
-    assert.equal(data.totalClaimedEth, mm.totalClaimedWei());
-  };
-
-  let waitForMonth = async function(numberOfMonth, startBlock, blocksPerPeriod) {
-    await help.waitToBlock(startBlock+blocksPerPeriod*numberOfMonth, accounts);
-  };
-
   it("Should return the correct sellPrice on every period with 873750 tokens total supply and 3285 ETH in MM", async function() {
     token = await simulateCrowdsale(100, [873750,0,0,0,0], accounts);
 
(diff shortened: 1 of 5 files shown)


test/MarketMaker.js
```diff
@@ -295,44 +303,57 @@ contract('marketMaker', function(accounts) {
...
-    checkScenarioProperties({
-      marketMakerEthBalance: 180, marketMakerLifBalance: 10, totalProfit: 0,
-      customerEthBalance: 520, customerLifBalance: 90,
-      marketMakerSellPrice: 2100, marketMakerBuyPrice: 2000,
+    await checkScenarioProperties({
+      marketMakerEthBalance: web3.toWei(180, 'ether'),
+      marketMakerLifBalance: help.lif2LifWei(10),
+      totalWeiProfit: 0,
+      customerEthBalance: customerInitEth.plus(web3.toWei(20)), customerLifBalance: 90,
+      marketMakerSellPrice: 2.1 * priceFactor, marketMakerBuyPrice: 2 * priceFactor,
```
kvbik: maybe if we do all the proper transformations in the checkScenarioProperties method, it will be bit more readable?

test/MarketMaker.js
```diff
@@ -295,44 +303,57 @@ contract('marketMaker', function(accounts) {
...
     // Month 0
     await waitForMonth(0, startBlock, blocksPerPeriod);
     // MMETH = 200,   TP = 0,    MMT = 0,   TC = 100, SP = 2100 mETH/Lif, BP = 2000, CL 0%,  maxClaimable = 0,   claimed = 0
-    checkScenarioProperties({
-      marketMakerEthBalance: 200, marketMakerLifBalance: 0, totalProfit: 0,
-      customerEthBalance: 500, customerLifBalance: 100,
-      marketMakerSellPrice: 2100, marketMakerBuyPrice: 2000,
+    await checkScenarioProperties({
+      marketMakerEthBalance: startingMMBalance, marketMakerLifBalance: 0, totalWeiProfit: 0,
+      customerEthBalance: customerInitEth, customerLifBalance: customerInitTokens,
```
kvbik: i am not sure, if doing the computations in the tests is not making it less readable. i've been playing with the idea and abandon it. you could end up doing (in the last par of the scenario) something like `startingMMBalance - 10 - 20 - 30 + 100`

kvbik: naming inconsistency: `startingMMBalance` vs `customerInitEth` isn't it basically the same?

test/MarketMaker.js
nberger: maybe. but for now I'm more worried in making it work

test/MarketMaker.js
nberger: We need to make it work first, then make it more readable. I mean, I don't want to make it unreadable but the most serious issue now is that the test doesn't run. For example now I'm getting an error "Cannot send value to non-payable constructor", but the constructor is marked as payable...

kvbik: sure thing. don't get me wrong. just wanted my comment to fall away somewhere. i want to make it work too!

nberger: I know. Sorry if it sounded harsh. It's just that I hate this new error :)

kvbik: haha, sure thing, i hate all of the errors that stay in the way!

nberger: It's weird because the constructor is marked as payable, I just checked `build/contracts/LifMarketMaker.json`, it has:

```
      "payable": true,
      "type": "constructor"
```

I think perhaps the constructor cannot be payable and receive the funds when constructed from another contract. This was working before today's update of truffle and solidity, but probably stopped working since that update.

@AugustoL any idea?

nberger: And even if it used to work on the previous truffle/solidity version, I'm not sure what would the syntax be to send the Ether when constructing the MarketMaker from LifCrowdsale. Maybe it will be easier to just have a separate `Fund() payable` function to send the funds, that can only be called once and do so from the LifCrowdsale

AugustoL: @nberger did you try removing the old compiled contracts and compiling the contracts again? Maybe you have an older version of the contract without the `payable` in the MM constructor

nberger: Thanks to @AugustoL, this error is fixed now: we were missing a parameter in the MarketMaker constructor call and that was causing the misleading error message

kvbik:
I was thinking the same way today. Like old migrations or something like
that. Cool!

test/MarketMaker.js
```diff
@@ -254,7 +236,30 @@ contract('marketMaker', function(accounts) {
     }
   });
 
-  it.only("should go through scenario with some claims and sells on the Market Maker", async function() {
+  var checkScenarioProperties = async function(data, mm, customer) {
```
nberger: good, we needed to move this block long ago. just one thing @AugustoL: next time you move code, try not to change it in the same commit, it makes it way harder to see what has changed. Try to do it in separate commits.

AugustoL: :+1:

nberger:
I'm going to rebase & squash some commits, then merge

PR #116, opened 2017-08-21
Change sellPrice calculation to be calculated on the MM contract

AugustoL:
Change sellPrice calculation to be calculated on the market maker contract and not take form mapping

### DIFF
--- a/contracts/LifMarketMaker.sol
+++ b/contracts/LifMarketMaker.sol
@@ -153,28 +153,18 @@ contract LifMarketMaker is Ownable {
     assert(totalPeriods == 24 || totalPeriods == 48);
     require(startBlock >= block.number);
     require(blocksPerPeriod > 0);
+    require(marketMakerPeriods[i].startBlock > 0);
 
     // The sellPriceIncrements represents how much is going to increase in % the sellPrice
     // every period.
 
-    uint256[48] memory accumSellPriceIncrements = [
-      uint256(0), 1000, 2010, 3030, 4060, 5101,
-      6152, 7213, 8285, 9368, 10462,
-      11566, 12682, 13809, 14947, 16096,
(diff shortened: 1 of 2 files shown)


contracts/LifMarketMaker.sol
```diff
@@ -153,28 +153,18 @@ contract LifMarketMaker is Ownable {
     assert(totalPeriods == 24 || totalPeriods == 48);
     require(startBlock >= block.number);
     require(blocksPerPeriod > 0);
+    require(marketMakerPeriods[i].startBlock > 0);
```
nberger: `i` is probably not defined here. Anyways, why not just `require(marketMakerPeriods.length > 0)`?

contracts/LifMarketMaker.sol
```diff
@@ -153,28 +153,18 @@ contract LifMarketMaker is Ownable {
...
-    for (uint8 i = 0; i < totalPeriods; i++) {
+    marketMakerPeriods[0].accumSellPriceIncrement = PRICE_FACTOR;
 
-      require(marketMakerPeriods[i].startBlock > 0);
+    for (uint8 i = 1; i < totalPeriods; i++) {
 
-      marketMakerPeriods[i].accumSellPriceIncrement = accumSellPriceIncrements[i];
+      marketMakerPeriods[i].accumSellPriceIncrement =
+        marketMakerPeriods[i-1].accumSellPriceIncrement
+        .mul(101000).div(PRICE_FACTOR);
```
nberger: if we are doing all of this, why don't we just store the price in the marketMakerPeriod, instead of the price increment? Do we use the increment for anything else?

AugustoL:
@nberger we can close this, since we are not going to use sellPrice on the market maker.

nberger:
@AugustoL agreed, closing now

PR #120, opened 2017-08-22
Using smaller balances on testrpc

nberger:


nberger:
Build is still failing so I'll close it: it's not critical and not worth it to spend more time on it

Issue #118, opened 2017-08-22
Integrate Crowdsale with Market Maker


nberger:
The Crowdsale should create the Market Maker when it finishes, with all the correct parameters and everything that's needed so it can operate correctly (create the periods, send the funds, transfer ownership to the foundation wallet, etc)

nberger:
Done in #125

PR #125, opened 2017-08-26
Integrate Market Maker in Crowdsale

AugustoL:
After the crowdsale ends if it succeed and raise more than 10M usd it iss going to create a market maker contract, the amount of periods of the market maker depends on the balance that it is going to receive, if it is less than 40M usd it will have 24 periods, if not 48.
To calculate the crowdsale caps and balance to be sent to the market maker with more precision,  the caps are set in USD dollars on the crowdsale contract and a function to set how much is the wei/USD before public presale and ICO has been added.

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -4,6 +4,7 @@ import "zeppelin-solidity/contracts/ownership/Ownable.sol";
 import "zeppelin-solidity/contracts/lifecycle/Pausable.sol";
 import "zeppelin-solidity/contracts/math/SafeMath.sol";
 import "./LifToken.sol";
+import "./LifMarketMaker.sol";
 
 contract LifCrowdsale is Ownable, Pausable {
   using SafeMath for uint256;
@@ -22,7 +23,24 @@ contract LifCrowdsale is Ownable, Pausable {
 
   // address where funds are collected
   address public foundationWallet;
-  address public marketMaker;
+
+  // minimun amount of wei to be raised in order to succed, it starts in USD
+  uint256 public maxPresaleCapUSD = 1000000;
+
+  // minimun amount of wei to be raised in order to succed, it starts in USD
+  uint256 public minCapUSD = 5000000;
+
+  // maximun balance that the foundation can have, it starts in USD
+  uint256 public maxFoundationCapUSD = 10000000;
+
+  // maximun balance that the 24 month market maker can have
+  uint256 public marketMaker24PeriodsCapUSD = 40000000;
+
+  // how much a USD worth in wei in public presale
+  uint256 public weiPerUSDinPresale = 0;
+
+  // how much a USD worth in wei in ICO
+  uint256 public weiPerUSDinICO = 0;
 
   // how much wei a token unit costs to a buyer, during the private presale stage
   uint256 public privatePresaleRate;
@@ -43,11 +61,8 @@ contract LifCrowdsale is Ownable, Pausable {
   // total amount of wei received as presale payments (both private and public)
   uint256 public totalPresaleWei;
 
-  // maximun amount of ether that can be raised using presale payments in wei unit
-  uint256 public maxPresaleWei;
-
-  //  minimun amount of wei to be raised in order to succed
-  uint256 public minCap;
+  // the address of teh market maker created at the end of the crowdsale
+  address public marketMaker;
 
   mapping(address => uint256) public purchases;
 
@@ -79,10 +94,7 @@ contract LifCrowdsale is Ownable, Pausable {
     uint256 _rate1,
     uint256 _rate2,
     uint256 _privatePresaleRate,
-    address _foundationWallet,
-    address _marketMaker,
-    uint256 _minCap,
-    uint256 _maxPresaleWei
+    address _foundationWallet
   ) {
     require(_publicPresaleStartBlock >= block.number);
     require(_publicPresaleEndBlock > _publicPresaleStartBlock);
@@ -92,9 +104,7 @@ contract LifCrowdsale is Ownable, Pausable {
     require(_publicPresaleRate > 0);
     require(_rate1 > 0);
     require(_rate2 > 0);
-    require(_minCap > 0);
     require(_foundationWallet != 0x0);
-    require(_marketMaker != 0x0);
 
     token = new LifToken();
     token.pause();
@@ -109,9 +119,20 @@ contract LifCrowdsale is Ownable, Pausable {
     rate2 = _rate2;
     privatePresaleRate = _privatePresaleRate;
     foundationWallet = _foundationWallet;
-    marketMaker = _marketMaker;
-    minCap = _minCap;
-    maxPresaleWei = _maxPresaleWei;
+  }
+
+  // Set how the rate wei per USD for the public presale, necesary to calculate with more
+  // precision the maxCap on the presale.
+  function setWeiPerUSDinPresale(uint256 _weiPerUSD) onlyOwner {
+    require (block.number < publicPresaleStartBlock && block.number > publicPresaleStartBlock.sub(10));
+    weiPerUSDinPresale = _weiPerUSD;
+  }
+
+  // Set how the rate wei per USD for the ICO, necesary to calculate with more precision the
+  // maxCap on the distribution of funds on finalize.
+  function setWeiPerUSDinICO(uint256 _weiPerUSD) onlyOwner {
+    require (block.number < startBlock && block.number > startBlock.sub(10));
+    weiPerUSDinICO = _weiPerUSD;
   }
 
   // returns the current rate or 0 if current block is not within the crowdsale period
@@ -142,6 +163,7 @@ contract LifCrowdsale is Ownable, Pausable {
   function buyTokens(address beneficiary) payable {
     require(beneficiary != 0x0);
     require(validPurchase());
+    assert(weiPerUSDinICO > 0);
 
     uint256 weiAmount = msg.value;
 
@@ -166,6 +188,7 @@ contract LifCrowdsale is Ownable, Pausable {
   function buyPresaleTokens(address beneficiary) payable {
     require(beneficiary != 0x0);
     require(validPresalePurchase());
+    assert(weiPerUSDinPresale > 0);
 
     uint256 weiAmount = msg.value;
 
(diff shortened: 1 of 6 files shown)


AugustoL:
fix #118

contracts/LifCrowdsale.sol
```diff
@@ -22,7 +23,21 @@ contract LifCrowdsale is Ownable, Pausable {
 
   // address where funds are collected
   address public foundationWallet;
-  address public marketMaker;
+
+  // minimun amount of wei to be raised in order to succed, it starts in USD
+  uint256 public maxCapPresaleUSD = 1000000;
```
nberger: maxPresaleCapUSD

contracts/LifCrowdsale.sol
```diff
@@ -22,7 +23,21 @@ contract LifCrowdsale is Ownable, Pausable {
...
+
+  // minimun amount of wei to be raised in order to succed, it starts in USD
+  uint256 public minCapUSD = 5000000;
+
+  // maximun balance that the foundation can have, it starts in USD
+  uint256 public foundationBalanceCapUSD = 10000000;
```
nberger: What about `minMarketMakerCapUSD`, so it's more consistent with the next variable?

contracts/LifCrowdsale.sol
```diff
@@ -22,7 +23,21 @@ contract LifCrowdsale is Ownable, Pausable {
...
+
+  // maximun balance that the 24 month market maker can have
+  uint256 public marketMaker24PeriodsCapUSD = 40000000;
+
+  // how much a USD worth in wei
+  uint256 public weiPerUSD = 0;
```
nberger: We need two of these variables: one for the public presale and one for the TGE

contracts/LifCrowdsale.sol
```diff
@@ -69,6 +78,12 @@ contract LifCrowdsale is Ownable, Pausable {
     uint256 amount
   );
 
+  /**
+   * event for market market creation logging
+   * @param contractAddr the contract address
+   */
+  event MMCreated(address contractAddr);
```
nberger: I'd not shorten the names here so I'd use `MarketMakerCreated` & `contractAddress`, what's the reason to shorten them?

contracts/LifCrowdsale.sol
```diff
@@ -109,9 +119,16 @@ contract LifCrowdsale is Ownable, Pausable {
...
     foundationWallet = _foundationWallet;
-    marketMaker = _marketMaker;
-    minCap = _minCap;
-    maxPresaleWei = _maxPresaleWei;
+  }
+
+  // Set how the rate wei per USD, necesary to calculate with more precision the
+  // maxCap on the presale and distribution of funds on finalize.
+  // This function can be called only 10 blocks before the publicPresaleStartBlock and startBlock
+  function setWeiPerUSD(uint256 _weiPerUSD) onlyOwner {
```
nberger: As I said above, I think we should use 2 separate variables for the presale and for the TGE. It will be more clear for everyone.

contracts/LifCrowdsale.sol
```diff
@@ -191,18 +210,41 @@ contract LifCrowdsale is Ownable, Pausable {
 
     uint256 tokens = weiSent.mul(privatePresaleRate);
 
-    require(totalPresaleWei.add(weiSent) <= maxPresaleWei);
```
nberger: Why is this validation gone? I understand that we need to modify it to take the wei/USD rate into account, but not remove it. Or am I missing something?

contracts/LifCrowdsale.sol
```diff
@@ -191,18 +210,41 @@ contract LifCrowdsale is Ownable, Pausable {
...
 
   // send ether to the fund collection wallet
-  function forwardFunds() onlyOwner {
-    foundationWallet.transfer(this.balance);
-    // TODO
-    // marketMaker.transfer(marketMakerPercentage * this.balance);
+  function forwardFunds() internal {
+
+    // calculate the max amount of wei for the foundation
+    uint256 foundationBalanceCapWei= foundationBalanceCapUSD.mul(weiPerUSD);
```
nberger: space before `=`

contracts/LifCrowdsale.sol
```diff
@@ -191,18 +210,41 @@ contract LifCrowdsale is Ownable, Pausable {
...
+
+    // if the minimiun cap for the market maker is not reached transfer all funds to foundation
+    // else if the min cap for the market maker is reached, create it and send the remaining funds
+    if (this.balance < foundationBalanceCapWei) {
```
nberger: as we are writing this `if` logic, why don't we add a condition for the TGE minimum balance being reached or not, with a TODO to return the contributed eth to the contributors?

contracts/LifCrowdsale.sol
```diff
@@ -191,18 +210,41 @@ contract LifCrowdsale is Ownable, Pausable {
...
+      uint256 mmFundBalance = this.balance.sub(foundationBalanceCapWei);
+
+      // check how much preiods we have to use on the market maker
+      uint8 marketMakerPeriods = 24;
+      if (mmFundBalance > marketMaker24PeriodsCapUSD.mul(weiPerUSD))
+        marketMakerPeriods = 48;
+
+      foundationWallet.transfer(foundationBalanceCapWei);
+      LifMarketMaker marketMaker = new LifMarketMaker(
+        address(token), block.number.add(10), 20, marketMakerPeriods, foundationWallet, 105000
```
nberger: we should add a variable for the delay between this finalize and the market maker startBlock. I see you are using `10` but we talked about using 2 weeks, remember? At least please add a TODO here

contracts/LifCrowdsale.sol
```diff
@@ -191,18 +210,41 @@ contract LifCrowdsale is Ownable, Pausable {
...
+      );
+      marketMaker.fund.value(mmFundBalance)();
+
+      MMCreated(marketMaker);
```
nberger: shouldn't we also have a `marketMaker` variable in the contract, that we set here, that gives an easy way to get to the marketMaker?

contracts/LifCrowdsale.sol
```diff
@@ -228,7 +270,8 @@ contract LifCrowdsale is Ownable, Pausable {
   }
 
   function funded() public constant returns (bool) {
-    return weiRaised >= minCap;
+    require(weiPerUSD > 0);
```
nberger: it should be `assert` instead of `require` because it's not a value provided by the user in this call

nberger: (the same applies to other similar `require`)

contracts/LifCrowdsale.sol
```diff
@@ -257,15 +300,18 @@ contract LifCrowdsale is Ownable, Pausable {
...
-
       token.transferOwnership(owner);
 
-    }
+      forwardFunds();
 
-    Finalized();
+    } else {
+
+      Finalized();
```
nberger: I think the event should trigger independently of whether it was successfully funded or not.

test/CrowdsaleGenTest.js
```diff
@@ -346,17 +399,21 @@ contract('LifCrowdsale Property-based test', function(accounts) {
   });
 
   let checkCrowdsaleState = async function(state, crowdsaleData, crowdsale) {
-    assert.equal(state.crowdsalePaused, await crowdsale.paused.call());
-    assert.equal(_.sumBy(state.purchases, (b) => b.tokens), help.lifWei2Lif(parseFloat(await crowdsale.tokensSold.call())));
+    assert.equal(state.crowdsalePaused, await crowdsale.paused());
+    assert.approximately(
+      _.sumBy(state.purchases, (b) => b.tokens),
+      parseFloat(help.lifWei2Lif(parseFloat(await crowdsale.tokensSold()))),
+      0.00001
```
nberger: Where does this difference come from? Also it would be nice to use a higher precision (more decimals, let's say 0.00000000001), or is this the lowest precision before the test breaks?

contracts/LifCrowdsale.sol
AugustoL: We really need two variables? I think one is enough, it can be changed X amount of blocks before publicPresale and ICO, on the TGE it will use the last value passed before the ICO.

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol
AugustoL: We dont need to access the marketMaker from the crowdsale, just create it, thats why I only added an event. I think we can store the address if we need to access it, but since we dont an event in enough, once the crowdsale contract ends it wont be used any more.

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol
nberger: I think it's better. It's definitely better to understand if they have clear names whether they refer to presale or TGE. Also imagine if the TGE starts right after the presale, then by updating the value for the TGE we are changing the value for the presale while it is running. I think the safest is to keep them separate

contracts/LifCrowdsale.sol
nberger: I mean users will know the address of the crowdsale stop they could go back to the crowdsale and take the market maker address from there. I'm not sure but it seems to me that extracting the value from the event is more complicated for users

contracts/LifCrowdsale.sol
AugustoL: Because we dont know the `weiPerUSDinPresale` yet, so we cannot check that the amount of tokens added on private presale are less that certain amount, but since the transactions with this functions would be recorded and the investors can see the totalSupply before the public presale we dont need to check the amount, I mean, if we issue a wrong number of tokens it would be there on the totalSupply.

contracts/LifCrowdsale.sol
AugustoL: I can rename it to `maxFoundationCapUSD` if you want, but I think its more important that we cant have more than 10M usd rather than the minimun balance for the MM is10M.

nberger:
Have you just pushed forced a rebased version of this? Please try to avoid that after someone started reviewing the PR and until the review is not finished: it's difficult what has been fixed/changed since the last review, I think it also marks all the comments as if they were on outdated diffs (and not only the ones where the code has changed), it makes reviewing really much harder than it can be.

So the idea is to only add new commits since the review has started or since someone fetched your changes (which can be as soon as you pushed) and then rebase/rewrite the commits only at the end, before merging

contracts/LifCrowdsale.sol
nberger: Ok, that's a good reason, let's use `maxFoundationCapUSD` as you suggest

contracts/LifCrowdsale.sol
```diff
@@ -43,11 +61,8 @@ contract LifCrowdsale is Ownable, Pausable {
   // total amount of wei received as presale payments (both private and public)
   uint256 public totalPresaleWei;
 
-  // maximun amount of ether that can be raised using presale payments in wei unit
-  uint256 public maxPresaleWei;
-
-  //  minimun amount of wei to be raised in order to succed
-  uint256 public minCap;
+  // the address of teh market maker created at the end of the crowdsale
```
nberger: s/teh/the

contracts/LifCrowdsale.sol
```diff
@@ -109,9 +119,20 @@ contract LifCrowdsale is Ownable, Pausable {
...
     foundationWallet = _foundationWallet;
-    marketMaker = _marketMaker;
-    minCap = _minCap;
-    maxPresaleWei = _maxPresaleWei;
+  }
+
+  // Set how the rate wei per USD for the public presale, necesary to calculate with more
+  // precision the maxCap on the presale.
+  function setWeiPerUSDinPresale(uint256 _weiPerUSD) onlyOwner {
+    require (block.number < publicPresaleStartBlock && block.number > publicPresaleStartBlock.sub(10));
```
nberger: So this leaves only the 10 blocks before the presale as the valid time to call this function? I think we should be able to call this function anytime until 10 minutes before the presale start time. The same applies to the next function of course

contracts/LifCrowdsale.sol
nberger: Ok, I agree this validation is not needed in the private presale. But I thought this was for the public presale, the diff confused me, github's fault ;) and now I see that this validation is in the `validPresalePurchase` so all good :+1:

contracts/LifCrowdsale.sol
nberger: thinking more about this, I see this is not needed because `forwardFunds` is only called when the minimum balance was reached

contracts/LifCrowdsale.sol
```diff
@@ -228,14 +276,15 @@ contract LifCrowdsale is Ownable, Pausable {
...
+    assert(weiPerUSDinICO > 0);
+    return weiRaised >= minCapUSD.mul(weiPerUSDinICO);
   }
 
   // return the eth if the crowdsale didnt reach the minCap
   function claimEth() public {
     require(isFinalized);
     require(hasEnded());
-    require(funded());
+    require(!funded());
```
nberger: Why is this changed to `!funded`? I think it should stay `funded`, eth can be claimed by the foundation only if the min cap was reached, right?

contracts/LifCrowdsale.sol
```diff
@@ -257,16 +306,16 @@ contract LifCrowdsale is Ownable, Pausable {
     // foward founds and unpause token only if minCap is reached
     if (funded()) {
 
+      // finish the minting of the token, unpause it and transfer teh ownership to the foundation
```
nberger: s/teh/the

contracts/LifCrowdsale.sol
AugustoL: This function is not for the foundation, its for the investors. If the minCap wasn't reached they can claim their ether back.

nberger: Oh, right! I was confusing the MarketMaker here :)
We have to add a claimEth command to the property-based test, it's obvious that this is not being tested because it should have made the tests to fail

AugustoL: :+1: going to create an issue for this

nberger:
LGTM :+1: 🎉  (aside from the conflicts and that you probably want to rebase/rewrite the commits)

PR #126, opened 2017-08-26
Pausable market maker

AugustoL:


contracts/LifMarketMaker.sol
```diff
@@ -298,4 +298,23 @@ contract LifMarketMaker is Ownable {
...
+  function pause() onlyOwner whenNotPaused {
+    paused = true;
+    blockPaused = block.number;
+    Pause();
+  }
+
+  function unpause() onlyOwner whenPaused {
+    uint256 blocksPaused = block.number.sub(blockPaused);
+    uint256 periodPaused = blockPaused.sub(startBlock).div(blocksPerPeriod);
+    marketMakerPeriods[periodPaused].endBlock = marketMakerPeriods[periodPaused].endBlock.add(blocksPaused);
```
nberger: Can we avoid updating all the blocks in unpause? Maybe keeping a general startBlock and then having every period's start and end blocks to be relative to that `startBlock.plus(blocksPaused)`?

AugustoL: For that we need to remove startBlock and endBlock form the periods structure, since we wont need it anymore. I think its a better solution, but we can leave it for the next PR.

nberger: I think we should do it in this PR, it'd be the best if we can do that:

- Initial commit with the change to having startBlock + relative blocks in the periods (as you said, probably removing start and end block from periods altogether)
- Second commit adding pausable

AugustoL: ok, will do

contracts/LifMarketMaker.sol
```diff
@@ -44,9 +44,16 @@ contract LifMarketMaker is Ownable {
 
   bool public funded = false;
 
+  // if the market maker is paused or not
+  bool public paused = false;
+
+  // total amount of blocks that the maret maker was paused
+  uint256 public totalBlocksPaused = 0;
+
+  // the block where the cmarket maker was paused
```
nberger: "market" has a typo here and in previous comment.
Also, what about changing `blockPaused` to `lastPausedBlock`? Or actually, why don't we change `paused` from bool to uint256 so it represents the block since which it's paused? We get rid of a var that way, we would set it back to 0 when it's not paused. What do you think?

AugustoL: I thought in that, but I think keeping that bool makes it more understandable, if you want to know if is paused or not as a dev you will look for a bool variable called `paused` and since its just a bool, its not expensive to keep it. 
About changing the name of `blockPaused` I can change it `pausedBlock` and `totalBlocksPaused` to `totalPausedBlocks`

AugustoL:
@nberger ready for final review, I rename the variables blockPaused and totalBlocksPaused

contracts/LifMarketMaker.sol
```diff
@@ -45,9 +45,16 @@ contract LifMarketMaker is Ownable {
   // Has the Market Maker been funded by calling `fund`? It can be funded only once
   bool public funded = false;
 
+  // if the market maker is paused or not
+  bool public paused = false;
+
+  // total amount of blocks that the maret maker was paused
```
nberger: please fix the typo here `maret` and in the next comment `cmarket`

contracts/LifMarketMaker.sol
```diff
@@ -197,7 +210,9 @@ contract LifMarketMaker is Ownable {
 
   // Called from the foundation wallet to claim eth back from the Market Maker. Maximum amount
   // that can be claimed is determined by getMaxClaimableWeiAmount
-  function claimEth(uint256 weiAmount) {
+
```
nberger: why this newline? Let's be consistent on this kind of things, after the comment before the function there's usually no newline, or is it? I also don't like the initial newline in the function body but I can live with it

contracts/LifMarketMaker.sol
```diff
@@ -128,17 +143,15 @@ contract LifMarketMaker is Ownable {
...
-        startBlockPeriod, startBlockPeriod.add(blocksPerPeriod).sub(1),
         deltaDistribution, accumDistribution
       ));
 
-      startBlockPeriod = startBlockPeriod.add(blocksPerPeriod);
     }
   }
 
   function getCurrentPeriodIndex() constant public returns(uint256) {
     require(block.number >= startBlock);
```
nberger: I think the require has to change too, it should be `block.number.sub(totalPausedBlocks) >= startBlock`
Also, it should be an `assert`

AugustoL: why? the call can be done at anytime after the marketMaker start, changing to asser yes, but I we dont need to change the condition

nberger: I think it's more correct to discount the totalPausedBlocks to assert the current block is after "the actual" market maker start block. But we could also remove this require and just rely on the next instruction to throw. Whatever works for me. But if we keep it, it should be changed to an assert

AugustoL: changed to assert

PR #130, opened 2017-08-27
Adding claimEth command to crowdsale gen test

### LINKED ISSUE #128: Add claimEth command for crowdsale prop-based-tests

nberger:
Fixes #128

### DIFF
--- a/test/CrowdsaleGenTest.js
+++ b/test/CrowdsaleGenTest.js
@@ -94,6 +94,12 @@ contract('LifCrowdsale Property-based test', function(accounts) {
     eth: jsc.nat(0,200)
   });
 
+  let claimEthCommandGen = jsc.record({
+    type: jsc.constant("claimEth"),
+    eth: jsc.nat(0, 200),
+    fromAccount: accountGen
+  });
+
   let runWaitBlockCommand = async (command, state) => {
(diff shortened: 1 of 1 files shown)


test/CrowdsaleGenTest.js
```diff
@@ -408,6 +415,32 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
+
+    let crowdsale = state.crowdsaleData,
+      { publicPresaleStartBlock, maxPresaleWei, privatePresaleRate } = crowdsale,
+      nextBlock = web3.eth.blockNumber + 1,
+      account = accounts[command.fromAccount],
+      purchases = _.filter(state.purchases, (p) => p.account == command.fromAccount);
+
+    let shouldThrow = !state.crowdsaleFinalized ||
+      !state.funded ||
+      (purchases.length == 0) ||
```
AugustoL: Do we need this condition really? if there are no purchases its not funded. right? thats what we expect.

nberger: that's what should happen, yes... but not bad to double check, there could be bugs on the funded state for example, or the other way around. For example there's one bug here: it should be `crowdsaleFunded` instead of `funded`

AugustoL: :+1:

PR #135, opened 2017-08-28
Fix tests

AugustoL:


test/LifToken.js
```diff
@@ -186,25 +186,24 @@ contract('LifToken', function(accounts) {
...
-
     await token.unpause({from: accounts[0]});
 
     // now burn should work
     await token.burn(burned, {from: accounts[5]});
 
     new BigNumber(initialBalance).minus(burned).
       should.be.bignumber.equal(await token.balanceOf(accounts[5]));
-    totalSupply0.minus(burned).should.be.bignumber.equal(await token.totalSupply.call());
+    totalSupply.minus(burned).should.be.bignumber.equal(await token.totalSupply.call());
```
nberger: I liked `totalSupply0`, it was like "the initial total supply", but it's ok, we can keep this change :)

PR #146, opened 2017-08-30
Vesting contract

AugustoL:
Add VestedPayment contract with tests and helpers to tests using timestamps.


contracts/VestedPayment.sol
```diff
@@ -0,0 +1,85 @@
...
+import "./LifToken.sol";
+
+contract VestedPayment is Ownable {
+  using SafeMath for uint256;
+
+  // when the vested schedule starts
+  uint256 public startTime;
+
+  // how much each period will last
+  uint256 public timePerPeriod;
```
nberger: what's the unit of this? I think it's seconds, it should be part of the name, and also in the comment

contracts/VestedPayment.sol
```diff
@@ -0,0 +1,85 @@
...
+
+import "zeppelin-solidity/contracts/math/SafeMath.sol";
+import "zeppelin-solidity/contracts/ownership/Ownable.sol";
+import "./LifToken.sol";
+
+contract VestedPayment is Ownable {
+  using SafeMath for uint256;
+
+  // when the vested schedule starts
+  uint256 public startTime;
```
nberger: what about `startTimestamp`?

contracts/VestedPayment.sol
```diff
@@ -0,0 +1,85 @@
...
+  // the amount of tokens to be vested in total
+  uint256 public tokens;
+
+  // how much tokens were claimed
+  uint256 public claimed;
+
+  // the token contract
+  LifToken public token;
+
+  // how much cliff period teh vesting schedule will have
```
nberger: I'd say it this way: "duration (in periods) of the initial cliff in the vesting schedule"

nberger: and the variable could be `cliffDuration`

contracts/VestedPayment.sol
```diff
@@ -0,0 +1,85 @@
...
+    uint256 period = block.timestamp.sub(startTime).div(timePerPeriod);
+
+    if ((period < cliffPeriods) || !funded) {
+      return 0;
+    } else if (period > totalPeriods) {
+      return tokens.sub(claimed);
+    } else if (period == 0) {
+      return tokens.div(totalPeriods).sub(claimed);
+    } else {
+      return tokens.div(totalPeriods).mul(period).sub(claimed);
```
nberger: I think the `mul(period)` should be `mul(period + 1)` and once that's done the previous if is not needed

contracts/VestedPayment.sol
```diff
@@ -0,0 +1,85 @@
...
+    totalPeriods = _totalPeriods;
+    cliffPeriods = _cliffPeriods;
+    token = LifToken(tokenAddress);
+  }
+
+  // fund the vesting contract with tokens allowed to be spent by the contract
+  function fund(uint256 _tokens) onlyOwner {
+    assert(!funded);
+    uint256 tokensAllowed = token.allowance(owner, address(this));
+    require(tokensAllowed >= _tokens);
```
nberger: we can remove the allowance check, the token takes care of that

test/VestedPayment.js
```diff
@@ -0,0 +1,249 @@
...
+var BigNumber = web3.BigNumber;
+
+const should = require('chai')
+  .use(require('chai-bignumber')(BigNumber))
+  .should();
+
+var VestedPayment = artifacts.require("./VestedPayment.sol");
+var LifToken = artifacts.require("./LifToken.sol");
+var LifCrowdsale = artifacts.require("./LifCrowdsale.sol");
+var LifToken = artifacts.require("./LifToken.sol");
```
nberger: duplicate LifToken require

test/helpers/increaseTime.js
```diff
@@ -0,0 +1,54 @@
...
+  increaseTimeTo: increaseTimeTo,
+  duration: {
+    seconds: function(val) { return val},
+    minutes: function(val) { return val * this.seconds(60) },
+    hours:   function(val) { return val * this.minutes(60) },
+    days:    function(val) { return val * this.hours(24) },
+    weeks:   function(val) { return val * this.days(7) },
+    years:   function(val) { return val * this.days(365)}
+  }
+}
```
nberger: missing newline

test/helpers/latestTime.js
```diff
@@ -0,0 +1,4 @@
+// Returns the time of the last mined block in seconds
+module.exports = function latestTime() {
+  return web3.eth.getBlock('latest').timestamp;
+}
```
nberger: missing newline

test/helpers/increaseTime.js
```diff
@@ -0,0 +1,54 @@
+var latestTime = require('./latestTime');
+
+// Increases testrpc time by the passed duration in seconds
+
+function increaseTime(duration) {
```
nberger: this and other functions here should make it clear in their names that they modify testrpc time. There is concept of time in javascript already and by default that's the time one talks about when talking about time in javascript

contracts/VestedPayment.sol
AugustoL: :+1:

contracts/VestedPayment.sol
AugustoL: Whats the difference? I mean..its a time I think this name is good as it is.

contracts/VestedPayment.sol
AugustoL: :+1:

contracts/VestedPayment.sol
AugustoL: :+1:

contracts/VestedPayment.sol
AugustoL: :+1:

contracts/VestedPayment.sol
nberger: `startTime` is usually about the time in the day, like "it starts at 10:00am", timestamp is more clear in that it's date and time. But it's ok, feel free to leave it as `startTime` if you prefer :)

test/helpers/increaseTime.js
AugustoL: :+1:

test/VestedPayment.js
```diff
@@ -0,0 +1,248 @@
...
+    assert.equal(accounts[2], await vestedPayment.owner.call());
+    assert.equal(true, await vestedPayment.funded.call());
+
+    // Go to period 1
+    await increaseTimeTestRPCTo(startTime+duration.days(30));
+
+    try {
+      await vestedPayment.claimTokens(help.lif2LifWei(15), {from: accounts[2]});
+    } catch (error) {
+      if (error.message.search('invalid opcode') == -1) throw error;
```
nberger: wanna replace this with `if (!help.isInvalidOpcodeEx(error)) throw error`? same for other occurrences in this file

contracts/VestedPayment.sol
```diff
@@ -0,0 +1,81 @@
...
+  // how much tokens are available to be claimed
+  function getAvailableTokens() public constant returns (uint256) {
+    uint256 period = block.timestamp.sub(startTime).div(secondsPerPeriod);
+
+    if ((period < cliffDuration) || !funded) {
+      return 0;
+    } else if (period > totalPeriods) {
+      return tokens.sub(claimed);
+    } else {
+      return tokens.div(totalPeriods).mul(period.add(1)).sub(claimed);
```
nberger: maybe first `mul` then `div`, for higher precision?

contracts/VestedPayment.sol
```diff
@@ -0,0 +1,81 @@
...
+    funded = true;
+  }
+
+  // how much tokens are available to be claimed
+  function getAvailableTokens() public constant returns (uint256) {
+    uint256 period = block.timestamp.sub(startTime).div(secondsPerPeriod);
+
+    if ((period < cliffDuration) || !funded) {
+      return 0;
+    } else if (period > totalPeriods) {
```
nberger: I think it should be `>=` here

test/VestedPayment.js
AugustoL: :+1:

contracts/VestedPayment.sol
AugustoL: It should, good catch.

contracts/VestedPayment.sol
AugustoL: :+1:

contracts/VestedPayment.sol
```diff
@@ -0,0 +1,81 @@
...
+
+  // duration (in periods) of the initial cliff in the vesting schedule
+  uint256 public cliffDuration;
+
+  // if the vested payment was funded or not
+  bool public funded = false;
+
+  function VestedPayment(
+    uint256 _startTime, uint256 _secondsPerPeriod,
+    uint256 _totalPeriods, uint256 _cliffDuration, address tokenAddress
```
nberger: should we `require(cliffDuration < totalPeriods)`? It avoids any confusion if somebody thinks the duration units is seconds when it's periods. I know it's explained in the `cliffDuration` comment but...

AugustoL: yes, we should. adding it.

PR #150, opened 2017-08-31
Rewrite README

### LINKED ISSUE #149: Rewrite the README
The README is super outdated, still talking about the crowdsale stages & DAO. We should update it with information about LifCrowdsale and LifMarketMaker, and no DAO

nberger:
Fixes #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,13 +3,12 @@
 Líf is the token of the Winding Tree platform.
 
 Líf is based on the ERC20 token protocol, with the option to also send information between token holders uint new methods transferData and transferDataFrom.
-Líf is also a DAO, the token holders can create proposals to change token variables and send value/data to another contracts.
 
 [![Build Status](https://travis-ci.org/windingtree/LifToken.svg?branch=master)](https://travis-ci.org/windingtree/LifToken)
 
 ## Requirements
 
-Node v7.6 or higher (versions before 7.6 do not support async/await)
+Node v7.6 or higher (versions before 7.6 do not support async/await that is used in the LifToken tests)
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,13 +3,12 @@
...
-Líf is also a DAO, the token holders can create proposals to change token variables and send value/data to another contracts.
 
 [![Build Status](https://travis-ci.org/windingtree/LifToken.svg?branch=master)](https://travis-ci.org/windingtree/LifToken)
 
 ## Requirements
 
-Node v7.6 or higher (versions before 7.6 do not support async/await)
+Node v7.6 or higher (versions before 7.6 do not support async/await that is used in the LifToken tests)
 
 ## Install
```
kvbik: what about a note about the optional `yarn install`? or we don't need it any longer?

README.md
```diff
@@ -36,100 +37,45 @@ npm install
...
+* To run a specific test: `npm test -- test/Crowdsale.js`
 
-Creates a new proposal on the token, the token holder that creates the proposal needs to have the more than the fee charged for the proposal creation.
-```
-newProposal(address target, uint value, string description, uint agePerBlock, bytes4 signature, bytes actionData)
-```
+There are also two environment variables (`GEN_TESTS_QTY` and `GEN_TESTS_TIMEOUT`) that regulate the duration/depth of the property-based tests, so for example:
 
-### Vote
+    GEN_TESTS_QTY=50 GEN_TESTS_TIMEOUT=300 npm test
```
kvbik: i think you can use \`\`\`sh for a syntax highlighting, well, it is not that more beautiful in this case:

```sh
  GEN_TESTS_QTY=50 GEN_TESTS_TIMEOUT=300 npm test
```

README.md
nberger: `yarn` is (not much but) a little more cumbersome than just `npm` so not requiring I think it's better. Maybe we could mention yarn as optional?

kvbik: i would love not to need anything else than `npm install`, just in my case i had to call `yarn`, so maybe we have to mention that. but again, would be great if not!

kvbik: ok, tried fresh `npm install` and `npm test` and everything worked smooth. no need for `yarn` in the documentation!

nberger: cool, merging then! :)

PR #152, opened 2017-08-31
Add pausable logic to transfer, approve and transferFrom

### LINKED ISSUE #104: Some checks of whether the token is paused are missing
The following functions are not checking whether the token is paused:

`approveData`, `transferData`, `transferDataFrom`

nberger:
Fixes #104

### DIFF
--- a/test/commands.js
+++ b/test/commands.js
@@ -375,6 +375,18 @@ let runTransferCommand = async (command, state) => {
   return state;
 }
 
+let getAllowance = (state, sender, from) => {
+  if (!state.allowances[sender])
+    state.allowances[sender] = {};
+  return state.allowances[sender][from] || 0;
+}
+
+let setAllowance = (state, sender, from, allowance) => {
+  if (!state.allowances[sender])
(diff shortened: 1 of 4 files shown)


test/commands.js
```diff
@@ -64,7 +70,7 @@ let runBuyTokensCommand = async (command, state) => {
     state.purchases = _.concat(state.purchases,
       {tokens: tokens, rate: rate, wei: weiCost, beneficiary: command.beneficiary, account: command.account}
     );
-    state.balances[command.beneficiary] = (state.balances[command.beneficiary] || 0) + tokens;
+    state.balances[command.beneficiary] = (state.balances[command.beneficiary] || 0) + help.lif2LifWei(tokens);
```
AugustoL: this is not a BigNumeber? if so `plus` ca be used instead of +

nberger: good catch, I'll fix it

nberger: Fixed in ee5f2e6

PR #155, opened 2017-09-01
Add doxygen-style comments

### LINKED ISSUE #95: Improve documentation in the contracts, use doxygen style
https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format and also check out contracts from other projects

nberger:
Fixes #95

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -6,79 +6,121 @@ import "zeppelin-solidity/contracts/math/SafeMath.sol";
 import "./LifToken.sol";
 import "./LifMarketValidationMechanism.sol";
 
+/**
+   @title Crowdsale for the Lif Token Generation Even
+
+   Implementation of the Lif Token Generation Event (TGE) Crowdsale: A 2 week
+   fixed price, uncapped
+   token sale, with a discounted rate during the 1st week and discounted rates
+   for contributions during the presale period and a Market Validation Mechanism
+   that will receive the funds over the USD 10M
+   soft cap.
+   The crowdsale has a minimum cap of USD 5M which in case of not being reached
+   by purchases made during the 2 week period the token will not start operating
+   and all funds sent during that period will be made available to be claimed by
+   the originating addresses.
+   Funds up to the USD 10M soft cap will be sent to the Winding Tree Foundation
+   wallet at the end of the crowdsale.
+   Funds over that amount will be put in a MarketValidationMechanism (MVM) smart
+   contract that guarantees a price floor for a period of 2 or 4 years, allowing
+   any token holder to burn their tokens in exchange of part of the eth amount
+   sent during the TGE in exchange of those tokens.
+ */
 contract LifCrowdsale is Ownable, Pausable {
   using SafeMath for uint256;
 
-  // The token being sold
+  // The token being sold.
   LifToken public token;
 
-  // start and end timestamps of the public presale
+  // Start and end timestamps of the public presale.
   uint256 public publicPresaleStartTimestamp;
   uint256 public publicPresaleEndTimestamp;
 
-  // start and end timestamps where investments are allowed (both inclusive)
+  // Beginning of the period where tokens can be purchased at rate `rate1`.
   uint256 public startTimestamp;
+  // Moment after which the rate to buy tokens goes from `rate1` to `rate2`.
   uint256 public end1Timestamp;
+  // Marks the end of the Token Generation Event.
   uint256 public end2Timestamp;
 
-  // address where funds are collected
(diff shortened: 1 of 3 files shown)


contracts/LifToken.sol
```diff
@@ -3,6 +3,13 @@ pragma solidity ^0.4.13;
 import "zeppelin-solidity/contracts/token/MintableToken.sol";
 import "zeppelin-solidity/contracts/lifecycle/Pausable.sol";
 
+/**
+ * @title Líf, the Winding Tree token
+ *
+ * Implementation of Líf, the ERC20 token for Winding Tree, with extra methods
+ * for transfers with contract calls.
```
AugustoL: what about "to transfer value and data allowing the execution of calls in transfer" ? or smth like that

contracts/LifToken.sol
```diff
@@ -17,20 +24,14 @@ contract LifToken is MintableToken, Pausable {
...
-
-  function transferFrom(address _from, address _to, uint256 _value) whenNotPaused returns (bool) {
-    return super.transferFrom(_from, _to, _value);
-  }
-
-  // approveData is an addition to ERC20 token methods. It allows approving the transference of value and execute a data call on the approval transaction
-  function approveData(address spender, uint value, bytes data) whenNotPaused {
+  /**
+   * @dev `approveData` is an addition to ERC20 token methods.
+   * It allows to approve the transfer of value and execute a call with data in the same transaction
```
AugustoL: It allows to approve the transfer of value and execute a call with the data sent

contracts/LifToken.sol
```diff
@@ -41,8 +42,14 @@ contract LifToken is MintableToken, Pausable {
 
   }
 
-  // transferData is an addition to ERC20 token methods. It allows to transfer value and data on each transaction
-  function transferData(address to, uint value, bytes data) whenNotPaused {
+  /**
+   * @dev Addition to ERC20 token methods. Transfer tokens to a specified address and make a contract call
```
AugustoL: Transfer tokens to a specified address and execute a call with the data sent

contracts/LifToken.sol
nberger: Even though I like "contract call" or "function call" more, I'm fine with "execution of calls". However, it should be singular like in "execution of a call on transfer", right? It's a single call, not multiple calls

AugustoL: yes, singular.

contracts/LifCrowdsale.sol
```diff
@@ -6,78 +6,119 @@ import "zeppelin-solidity/contracts/math/SafeMath.sol";
...
+ */
 contract LifCrowdsale is Ownable, Pausable {
   using SafeMath for uint256;
 
-  // The token being sold
+  // The token being sold.
   LifToken public token;
 
-  // start and end timestamps of the public presale
+  // start and end timestamps of the public presale.
```
AugustoL: It shouldn't start with uppercase here like all the comments?

AugustoL: It shouldn't start with uppercase here like all the comments?

contracts/LifCrowdsale.sol
```diff
@@ -6,78 +6,119 @@ import "zeppelin-solidity/contracts/math/SafeMath.sol";
...
-  // start and end timestamps where investments are allowed (both inclusive)
+  // Beginning of the period where tokens can be purchased at rate `rate1`.
   uint256 public startTimestamp;
+  // Moment after which the rate to buy tokens goes from `rate1` to `rate2`.
   uint256 public end1Timestamp;
+  // Marks the end of the Token Generation Event.
   uint256 public end2Timestamp;
 
-  // address where funds are collected
+  // address of the Winding Tree Foundation wallet. Funds up to the soft cap are
```
AugustoL: start with uppercase here too?

nberger:
What style of block do we like the most, this one:

```
  /**
   * @dev Addition to ERC20 token methods. Transfer tokens from one address to another and make a contract call
   * on the same transaction
   *
   * @param from The address which you want to send tokens from
   * @param to The address which you want to transfer to
   * @param value The amout of tokens to be transferred
   * @param data ABI-encoded contract call. For example generated using web3's getData method
   */
```

Or this one:
```
  /**
     @dev Addition to ERC20 token methods. Transfer tokens from one address to another and make a contract call
     on the same transaction

     @param from The address which you want to send tokens from
     @param to The address which you want to transfer to
     @param value The amout of tokens to be transferred
     @param data ABI-encoded contract call. For example generated using web3's getData method
   */
```

I like the second better

AugustoL:
The second :)

PR #157, opened 2017-09-02
Add timestamps on LifCrowdsale

AugustoL:
* Change LifCrowdsale contract to work only with timestamps.
* Change the prop based tests for LifCrowdsale to work with timestamps.
* Using same simulateCrowdsale function form helpers on all tests.

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -12,14 +12,14 @@ contract LifCrowdsale is Ownable, Pausable {
   // The token being sold
   LifToken public token;
 
-  // start and end of the public presale
-  uint256 public publicPresaleStartBlock;
-  uint256 public publicPresaleEndBlock;
+  // start and end timestamps of the public presale
+  uint256 public publicPresaleStartTimestamp;
+  uint256 public publicPresaleEndTimestamp;
 
-  // start and end block where investments are allowed (both inclusive)
-  uint256 public startBlock;
-  uint256 public endBlock1;
-  uint256 public endBlock2;
+  // start and end timestamps where investments are allowed (both inclusive)
+  uint256 public startTimestamp;
+  uint256 public end1Timestamp;
+  uint256 public end2Timestamp;
 
   // address where funds are collected
   address public foundationWallet;
@@ -42,8 +42,8 @@ contract LifCrowdsale is Ownable, Pausable {
   // how much a USD worth in wei in TGE
   uint256 public weiPerUSDinTGE = 0;
 
-  // amount of blocks where the weiPerUSD cannot change before setWeiPerUSD functions
-  uint256 public setWeiLockBlocks = 0;
+  // amount of seconds where the weiPerUSD cannot change before setWeiPerUSD functions
+  uint256 public setWeiLockSeconds = 0;
 
   // how much wei a token unit costs to a buyer, during the private presale stage
   uint256 public privatePresaleRate;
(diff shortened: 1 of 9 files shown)


contracts/LifCrowdsale.sol
```diff
@@ -13,13 +13,13 @@ contract LifCrowdsale is Ownable, Pausable {
   LifToken public token;
 
   // start and end of the public presale
-  uint256 public publicPresaleStartBlock;
-  uint256 public publicPresaleEndBlock;
+  uint256 public publicPresaleStartTime;
```
nberger: I think `Time` as suffix is misleading here. It sounds like the time of the day, but of course it's a timestamp, meaning the date and the time. Timestamp as suffix would be too long so what if we use `At`? `publicPresaleStartAt`, `publicPresaleEndAt`, `startAt`, `end1At`, `end2At` etc.

What do you think?

contracts/LifCrowdsale.sol
```diff
@@ -13,13 +13,13 @@ contract LifCrowdsale is Ownable, Pausable {
...
+  uint256 public publicPresaleEndTime;
 
   // start and end block where investments are allowed (both inclusive)
```
nberger: we need to update the comments that refer to blocks, this one and the others

contracts/LifCrowdsale.sol
```diff
@@ -265,24 +266,24 @@ contract LifCrowdsale is Ownable, Pausable {
...
+    bool withinPeriod = currentTime >= startTime && currentTime <= endTime2;
     bool nonZeroPurchase = msg.value != 0;
     return (withinPeriod && nonZeroPurchase);
   }
 
   // @return true if the transaction can buy tokens on presale
   function validPresalePurchase() internal constant returns (bool) {
-    uint256 current = block.number;
-    bool withinPublicPresalePeriod = current >= publicPresaleStartBlock && current <= publicPresaleEndBlock;
+    uint256 currentTime = block.timestamp;
```
nberger: I'm not a fan of the name of this var, I'd use `now` instead or just remove it and use `block.timestamp` right away

test/CrowdsaleGenTest.js
```diff
@@ -174,44 +177,45 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
       commands: [
-        {"type":"waitBlock","blocks":10},
-        {"type":"sendTransaction","account":3,"beneficiary":0,"eth":9}
+        { type: "waitTime","seconds":duration.days(1)},
+        { type:"sendTransaction","account":3,"beneficiary":0,"eth":9}
       ],
       crowdsale: {
         publicPresaleRate: 33, rate1: 18, rate2: 33, privatePresaleRate: 48,
-        foundationWallet: 1, setWeiLockBlocks: 1, owner: 7
+        foundationWallet: 1, setWeiLockBlocks: 600, owner: 7
```
nberger: rename to `setWeiLockSeconds`

test/CrowdsaleGenTest.js
```diff
@@ -174,44 +177,45 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
-  it("calculates correct rate on the boundaries between endBlock1 and endBlock2", async function() {
+  it("calculates correct rate on the boundaries between endTime1 and endTime2", async function() {
     let crowdsaleAndCommands = {
       commands: [
-        { type: 'checkRate' },
-        { type: 'setWeiPerUSDinPresale', wei: 3000000000000000, fromAccount: 3 },
-        { type: 'checkRate' },
-        { type: 'waitBlock', blocks: 29 },
-        { type: 'buyTokens', beneficiary: 3, account: 2, eth: 12 }
+        { type: "checkRate" },
```
nberger: why all the changes from single quote ' to double quotes "? Does your editor does it automatically? we have a big diff already to add all those :)

test/CrowdsaleGenTest.js
```diff
@@ -295,9 +298,9 @@ contract('LifCrowdsale Property-based test', function(accounts) {
 
     await runGeneratedCrowdsaleAndCommands({
       commands: [
-        {"type":"setWeiPerUSDinTGE","wei":0,"fromAccount":10},
-        {"type":"setWeiPerUSDinTGE","wei":0,"fromAccount":6},
-        {"type":"setWeiPerUSDinTGE","wei":3,"fromAccount":6}
+        { type:"setWeiPerUSDinTGE","wei":0,"fromAccount":10},
```
nberger: why removing the double quotes to `type`, but not to the others? better not change any of them I guess :)

test/MarketMaker.js
```diff
@@ -10,6 +10,9 @@ var LifMarketMaker = artifacts.require("./LifMarketMaker.sol");
 var LifToken = artifacts.require("./LifToken.sol");
 var LifCrowdsale = artifacts.require("./LifCrowdsale.sol");
 
+var latestTime = require('./helpers/latestTime');
+var {increaseTimeTestRPC, increaseTimeTestRPCTo, duration} = require('./helpers/increaseTime');
```
nberger: Are these two requires needed here? I don't think so

contracts/LifCrowdsale.sol
kvbik: if we are consistent about that, why not..

contracts/LifCrowdsale.sol
AugustoL: we can use `now` is a global var that refers to block.timestamp

test/CrowdsaleGenTest.js
AugustoL: I changed the all to use single quotes.

test/CrowdsaleGenTest.js
AugustoL: I will remove the quotes on all properties.

test/MarketMaker.js
AugustoL: They dont, removing them.

contracts/LifCrowdsale.sol
AugustoL: I prefer using `time` on the name instead of `at`. because it can be at a block number too. For me time its ok, there is two ways to verify "time" conditions on solidity, blocks or timestamps, and I think using time is enough and doesnt bring any misleading.

AugustoL: And we are already using `time` on VestedPayment contract.

test/CrowdsaleGenTest.js
nberger: But try not to do it in this PR, it's already quite big

AugustoL: ok, do you want me to remove all quotes changes too?

contracts/LifCrowdsale.sol
nberger: I know we are using time in vested payment, I suggested to change it to timestamp, you preferred to keep it and I said "ok" because it was just one variable. Now it's a more widespread thing. Think why it's `block.timestamp` instead of `block.time`. We should at least be consistent with that, again I think time is more about time of the day, not an exact moment in time

contracts/LifCrowdsale.sol
nberger: 👍🏼

nberger: 👍🏼

test/CrowdsaleGenTest.js
nberger: I don't know, if it takes too much time just keep it, if you can easily revert just those changes that's better. Try not to introduce them the next time :)

AugustoL: I will remove them when I re order the commits once its approved.

test/CrowdsaleGenTest.js
nberger: I don't see them changed to single quotes

AugustoL: I will remove the changes on this lines on the diff, I can do all that changes later.

contracts/LifCrowdsale.sol
```diff
@@ -12,14 +12,14 @@ contract LifCrowdsale is Ownable, Pausable {
...
+  uint256 public publicPresaleStartTimestamp;
+  uint256 public publicPresaleEndTimestamp;
 
-  // start and end block where investments are allowed (both inclusive)
-  uint256 public startBlock;
-  uint256 public endBlock1;
-  uint256 public endBlock2;
+  // start and end timestamps where investments are allowed (both inclusive)
+  uint256 public startTimestamp;
+  uint256 public endTimestamp1;
```
nberger: I think it should be `end1Timestamp`, the same for end2

PR #158, opened 2017-09-04
Add property-based test command to send tokens to MarketMaker

nberger:
It's the first step in adding the Market Maker to the property-based test.

### DIFF
--- a/test/MarketMaker.js
+++ b/test/MarketMaker.js
@@ -200,9 +200,9 @@ contract('marketMaker', function(accounts) {
     data.ethBalances[customerAddressIndex].should.be.bignumber.equal(web3.eth.getBalance(customer));
     data.balances[customerAddressIndex].should.be.bignumber.equal(await token.balanceOf(customer));
 
-    data.marketMakerMaxClaimableEth.should.be.bignumber.equal(await mm.getMaxClaimableWeiAmount());
+    data.marketMakerMaxClaimableWei.should.be.bignumber.equal(await mm.getMaxClaimableWeiAmount());
 
-    data.marketMakerClaimedEth.should.be.bignumber.equal(await mm.totalWeiClaimed.call());
+    data.marketMakerClaimedWei.should.be.bignumber.equal(await mm.totalWeiClaimed.call());
   };
 
   it("should go through scenario with some claims and sells on the Market Maker", async function() {
@@ -230,8 +230,8 @@ contract('marketMaker', function(accounts) {
       ethBalances: {},
       balances: {},
       marketMakerBuyPrice: startingMMBalance.dividedBy(help.lif2LifWei(tokenTotalSupply)).mul(priceFactor),
-      claimablePercentage: 0, marketMakerMaxClaimableEth: new BigNumber(0),
-      marketMakerClaimedEth: new BigNumber(0)
+      claimablePercentage: 0, marketMakerMaxClaimableWei: new BigNumber(0),
+      marketMakerClaimedWei: new BigNumber(0)
     };
     state.ethBalances[customerAddressIndex] = web3.eth.getBalance(customer);
     state.balances[customerAddressIndex] = await token.balanceOf(customer);
@@ -256,12 +256,12 @@ contract('marketMaker', function(accounts) {
       6583, 7243, 7929, 8640, 9377, 10138
     ];
 
-    let getMaxClaimableEth = function(state) {
+    let getMaxClaimableWei = function(state) {
       if (state.marketMakerMonth >= periods) {
-        help.debug("calculating maxClaimableEth with", startingMMBalance, state.marketMakerClaimedEth,
+        help.debug("calculating maxClaimableEth with", startingMMBalance, state.marketMakerClaimedWei,
           state.returnedWeiForBurnedTokens);
         return startingMMBalance.
-          minus(state.marketMakerClaimedEth).
+          minus(state.marketMakerClaimedWei).
           minus(state.returnedWeiForBurnedTokens);
       } else {
         const totalSupplyWei = web3.toWei(tokenTotalSupply, 'ether');
@@ -269,7 +269,7 @@ contract('marketMaker', function(accounts) {
           mul(state.claimablePercentage).dividedBy(priceFactor).
           mul(totalSupplyWei - state.marketMakerBurnedTokens).
           dividedBy(totalSupplyWei).
-          minus(state.marketMakerClaimedEth);
+          minus(state.marketMakerClaimedWei);
         return _.max([0, maxClaimable]);
       }
     }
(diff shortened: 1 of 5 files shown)


test/commands.js
```diff
@@ -437,6 +439,78 @@ let runTransferFromCommand = async (command, state) => {
...
+
+// TODO: implement finished
+let isMarketMakerFinished = (state) => false
+
+let runMarketMakerSendTokensCommand = async (command, state) => {
+  if (state.marketMaker === undefined) {
+    // doesn't make sense to execute the actual command, let's just assert
+    // that the crowdsale was not funded (in which case there should be MM)
+    assert.equal(false, state.crowdsaleFinalized && state.crowdsaleFunded,
+      "if there's no market Maker, crowdsale should not have been funded");
```
AugustoL: we can reach the minCap of 5M usd but not the 10M usd needed for the MM creation

test/MarketMaker.js
```diff
@@ -196,55 +195,65 @@ contract('marketMaker', function(accounts) {
...
     const initialBuyPrice = startingMMBalance.dividedBy(help.lif2LifWei(tokenTotalSupply));
 
     let state = {
-      month: 0,
-      burnedTokens: new BigNumber(0),
+      marketMakerMonth: 0,
+      marketMakerPeriods: periods,
+      token: token,
+      initialTokenSupply: help.lif2LifWei(tokenTotalSupply),
+      marketMakerBurnedTokens: new BigNumber(0), // burned tokens in MM, via sendTokens txs
```
AugustoL: do we need this? why is important to keep track of the burned tokens on the MM?

test/MarketMaker.js
```diff
@@ -265,19 +276,19 @@ contract('marketMaker', function(accounts) {
...
         return startingMMBalance.
-          minus(state.totalClaimedEth).
+          minus(state.marketMakerClaimedEth).
           minus(state.returnedWeiForBurnedTokens);
       } else {
         const totalSupplyWei = web3.toWei(tokenTotalSupply, 'ether');
         const maxClaimable = startingMMBalance.
           mul(state.claimablePercentage).dividedBy(priceFactor).
-          mul(totalSupplyWei - state.burnedTokens).
+          mul(totalSupplyWei - state.marketMakerBurnedTokens).
```
AugustoL: `.minus` form BigNumber can be used here

test/commands.js
```diff
@@ -437,6 +439,78 @@ let runTransferFromCommand = async (command, state) => {
...
+
+    try {
+      help.debug('Selling ',command.tokens, ' tokens in exchange of ', web3.fromWei(tokensCost, 'ether'), 'eth');
+      tx1 = await state.token.approve(state.marketMaker.address, lifWei, {from: fromAddress}),
+        tx2 = await state.marketMaker.sendTokens(lifWei, {from: fromAddress}),
+        gas = tx1.receipt.gasUsed + tx2.receipt.gasUsed;
+
+      help.debug("sold tokens to market Maker");
+
+      state.ethBalances[command.from] = initialEthBalance.plus(tokensCost).minus(help.gasPrice.mul(gas));
```
AugustoL: why `initialEthBalance` is used here?

test/MarketMaker.js
```diff
@@ -196,55 +195,65 @@ contract('marketMaker', function(accounts) {
...
-    assert.equal(data.month >= periods, await mm.isFinished());
+    assert.equal(data.marketMakerMonth >= periods, await mm.isFinished());
 
-    data.customerEthBalance.should.be.bignumber.equal(web3.eth.getBalance(customer));
-    data.customerLifBalance.should.be.bignumber.equal(await token.balanceOf(customer));
+    data.ethBalances[customerAddressIndex].should.be.bignumber.equal(web3.eth.getBalance(customer));
+    data.balances[customerAddressIndex].should.be.bignumber.equal(await token.balanceOf(customer));
 
-    data.maxClaimableEth.should.be.bignumber.equal(await mm.getMaxClaimableWeiAmount());
+    data.marketMakerMaxClaimableEth.should.be.bignumber.equal(await mm.getMaxClaimableWeiAmount());
```
AugustoL: this shouldnt be `marketMakerMaxClaimableWei` instead of `marketMakerMaxClaimableEth` ? we are comparing it again a wei amount

test/MarketMaker.js
```diff
@@ -196,55 +195,65 @@ contract('marketMaker', function(accounts) {
...
 
-    data.totalClaimedEth.should.be.bignumber.equal(await mm.totalWeiClaimed.call());
+    data.marketMakerClaimedEth.should.be.bignumber.equal(await mm.totalWeiClaimed.call());
```
AugustoL: Here to, the var name is on eth unit, buts its in wei.

test/MarketMaker.js
```diff
@@ -196,55 +195,65 @@ contract('marketMaker', function(accounts) {
...
   };
 
-  it("should go through scenario with some claims and sells on the Market Maker", async function() {
+  it.only("should go through scenario with some claims and sells on the Market Maker", async function() {
```
AugustoL: remove `.only`

test/MarketMaker.js
nberger: in the prop-based test we want to keep track of tokens burned in MM and tokens burned in general separately, to assert on both different numbers. We don't really need to keep track of both in this test, but we are migrating to using the commands from the prop-based test which have to update everything for that test, so here we are just adding both to be compatible with those commands

test/MarketMaker.js
nberger: good catch, fixed in 2e49ff2

test/commands.js
nberger: good catch, I'll fix it

test/commands.js
nberger: good question! haha
it sounds misleading but I think it's correct: it's the balance of the `command.from` account before doing the `approve` + `sendTokens`, it's initialized in this function. Maybe it sounded like it's the initial balance when the Market Maker was created or something like that?

test/commands.js
nberger: by the way, one problem we are having with adding property based tests for the market maker is that the sequence of commands usually make it so the market maker is not created, even the crowdsale is not finished and funded, so in practical terms this is the branch that's executed 99.9999% of the times. It'd be nice to find a way to have the crowdsale to finish funded more frequently in the tests

nberger: Expanded the comment in 23bc842, not the full solution but good enough for now?

nberger:
@augustol I think I addressed all your comments, could you please review again?

test/commands.js
AugustoL: Yes, that was I understood when I was reading it, we can change it fro `initialSenderWeiBalance` or something like that.

test/MarketMaker.js
AugustoL: cant we use the same value value for both and check that the amount of tokens burned match before and after we burned them outside and inside MM? We dont have a variable on the MM contract that keeps track of the tokens burned, thats why I think we dont need it and it add more complexity to what we want to tests that is just token supply. In fact, we can use only totalSupply that is the only variable we are using to keep track of the total tokens amount.

test/MarketMaker.js
nberger: I changed it back to use `-` instead of `.minus`, because `totalSupplyWei` was not a bigNumber so the test was failing... we could make it a BigNumber, but in reality we know it is small enough that we can keep it as is (it's 100 lifWei minus the burnedtokens which is also a relatively small number)

PR #162, opened 2017-09-05
Add solidity-coverage & coveralls CI reporting

cgewecke:
@AugustoL and @kvbik suggested this PR might be useful 😄 

It: 
+ installs `solidity-coverage` and `coveralls` (using `yarn add ... --dev`).
+ adds some convenience scripts to enable coverage generation locally and in Travis.
+ modifies `travis.yml` to generate coverage in an `after_script` and send the data to Coveralls. 
+ Modifies `travis.yml` to use `sudo: required` (increases the container's memory cap.)
+ adds about 5 min. to the total CI run time.

Note: The `after_script` does not separate the tests like the `script` in Travis because the results need to be aggregated. Not sure if that matters or not. 

You can turn Coveralls on as a GitHub integration by opening an account with them (it's free) and toggling this repo on. There's no CI token - it should just work the next time you build. There's a sample Coveralls report from my fork of `LifToken` [here](https://coveralls.io/github/cgewecke/LifToken).

Occasionally solidity-coverage's testrpc crashes with `Key not found` or `Blocklogs index etc` If you hit rebuild on the CI container it will usually generate a clean test run. 

LifToken's test coverage is really good. Worth noting that because there's an element of randomness in your suite, readings might vary a little from build to build. 

liftoken

If you have any problems with this, need changes or it breaks, please just let me know. I'm happy to fix / maintain it. Thanks!


cgewecke:
I guess this build will fail until Coveralls is turned on.  If you click on the after_script link in the Travis output you can see the report generated and get kicked back with a bad http response when it tries to run the hook.

AugustoL:
@cgewecke Thanks! Im enabling coveralls on the LifToken repo now, lets see if it works :)

I think we can also add the coverage badge to the README, for markdown it will be like this;

[Coverage Status](https://coveralls.io/github/windingtree/LifToken?branch=master)

.gitignore
```diff
@@ -2,3 +2,7 @@ node_modules
 build
 npm-debug.log
 .node-xmlhttprequest-sync-[0-9]*
+coverage
+coverage.json
+allFiredEvents
+scTopics
```
nberger: please add newlines at the end of files, there's a few missing like this

.travis.yml
```diff
@@ -1,5 +1,5 @@
 dist: trusty
-sudo: false
+sudo: required
```
nberger: why is sudo required?

.travis.yml
```diff
@@ -13,3 +13,6 @@ script:
   - jshint --show-non-errors test/*.js || echo "Errors in jshint ignored bc it doesn't support async/await"
   - GEN_TESTS_QTY=20 yarn test test/LifToken.js test/Crowdsale.js test/CrowdsaleGenTest.js
   - yarn test test/MarketMaker.js
+after_script:
+  - chmod +x scripts/coveralls.sh
```
nberger: I think you could add this change on the actual file in this PR so this line would not be needed. Is there any downside?

scripts/coverage.sh
```diff
@@ -0,0 +1,43 @@
...
+    --account="0x38062255973f02f1b320d8c7762dd286946b3e366f73076995dc859a6346c2ec,10000000000000000000000000000000000000000000000000000000000000000000000000000000" \
+    --account="0x35b5042e809eab0db3252bad02b67436f64453072128ee91c1d4605de70b27c1,10000000000000000000000000000000000000000000000000000000000000000000000000000000" \
+  > /dev/null &
+  testrpc_pid=$!
+fi
+
+SOLIDITY_COVERAGE=true GEN_TESTS_QTY=20 ./node_modules/.bin/solidity-coverage
+
+test_result=$?
+
```
nberger: shouldn't we kill testrpc before exit, if we started it? That's what we are doing in `scripts/test.sh`, we are not using long-running testrpc instances. It starts quite fast and tests are not prepared to run multiple times at the same time so we prefer to start from a clean slate. What do you think?

nberger:
This looks great overall, thank you @cgewecke! Even though I'm not a _huge_ fan of coverage, it's nice to have it. And great to hear your comment about the current coverage, we are putting much effort on that and even then there's still room for improvement.

One thing that worries me is the extra 5 minutes in build time. I guess most of the time is gone running the tests, right? What can we do then? Maybe run the coverage in master only?

nberger:
By the way, if we wanted to run the coverage in master only we'll probably have to use the [travis environment variables](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) like `TRAVIS_PULL_REQUEST`, `TRAVIS_BRANCH`, etc

AugustoL:
Agree with @nberger we can use it only on master.

scripts/coverage.sh
cgewecke: @nberger Line 4 of the script should run the cleanup routine on exit and shut testrpc down. However I've removed the logic about exiting with the test result in my revisions because it's confusing and redundant here.

.travis.yml
cgewecke: Cool! I didn't know you can do that - I thought this was something weird about yarn. Thanks!

.travis.yml
cgewecke: This is 'preventative' - it increases the container RAM from 4GB to 7.5GB. In the past I've seen problems with the Travis memcap when running multiple instances of testrpc  because there's very poor garbage collection there. And solidity-coverage has a relatively large memory footprint. It *may* not be necessary - I think the only drawback is that it takes slightly longer for the container to boot. [Info about this here.](https://docs.travis-ci.com/user/reference/overview/). Let me know if you think this should be undone.

scripts/coverage.sh
nberger: Oh right, and then there's the `trap`, cool :+1:

.travis.yml
nberger: Nice, thanks for adding that info, I didn't know about this. It might explain why our builds fail sporadically with issues connecting to `testrpc` which was probably because of OOM. Let's keep it :+1:

cgewecke:
@AugustoL @nberger Agree! You might also be interested in looking a Francisco Giordano's recent solution to this issue at Zeppelin which involves parallelizing the Travis build. These are his [`.yml`](https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/.travis.yml) and [`test.sh`](https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/scripts/test.sh).

@AugustoL Should I also push the badge in?

nberger:
Cool! Feel free to add the badge, or we can do it later. The same for running coverage in master only, someone can do it in a further PR. Again, feel free to add those changes but this is good enough to me :)

cgewecke:
Great. I'll add the badge right now but I'm going to let you do the master only coverage run. And learn by watching. 😄

PR #164, opened 2017-09-05
Run all the tests in travis from a single command

nberger:
We were running MarketMaker separately because there was a kind of random error
that was making the tests to fail sometimes and having it separate made it better.
But then we updated a few things (zeppelin, testrpc) and hopefully those random errors are gone

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,5 @@ cache:
 script:
   - set -e
   - jshint --show-non-errors test/*.js || echo "Errors in jshint ignored bc it doesn't support async/await"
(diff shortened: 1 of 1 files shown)


nberger:
Let's keep running the tests separately because it seems to help avoiding testrpc crashes. Other projects are doing the same: https://github.com/gnosis/MultiSigWallet/commit/de2769427f4bd2f90cd13fd9d4ca6eedd994c225

Even though we get some of those crashes like in this build: https://travis-ci.org/windingtree/LifToken/builds/273630681, at least is not that frequent as before.

I'm actually thinking we should run `test/CrowdsaleGenTest.js` in a third test process, might send a PR to do that soon

PR #166, opened 2017-09-06
Add zero address to accountGen, to force exceptions in txs in tests

### LINKED ISSUE #87: Add tests with address 0, to verify that the txs throw
We could add it to the accountGen so it could generate address==0 and then we should assert that the actual txs are throwing correctly.

nberger:
Transactions should throw when address == 0. Property-based test commands should handle the exceptions fine

Fixes #87

### DIFF
--- a/test/CrowdsaleGenTest.js
+++ b/test/CrowdsaleGenTest.js
@@ -64,7 +64,8 @@ contract('LifCrowdsale Property-based test', function(accounts) {
     help.debug("crowdsaleTestInput data:\n", input, publicPresaleStartTimestamp, publicPresaleEndTimestamp, startTimestamp, end1Timestamp, end2Timestamp);
 
     let {publicPresaleRate, rate1, rate2, owner, setWeiLockSeconds} = input.crowdsale,
-      ownerAddress = accounts[input.crowdsale.owner];
+      ownerAddress = gen.getAccount(input.crowdsale.owner),
+      foundationWallet = gen.getAccount(input.crowdsale.foundationWallet);
     let shouldThrow = (publicPresaleRate == 0) ||
       (rate1 == 0) ||
       (rate2 == 0) ||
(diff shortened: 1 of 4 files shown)


test/generators.js
```diff
@@ -3,12 +3,24 @@ var jsc = require("jsverify");
 // this is just to have web3 available and correctly initialized
 artifacts.require("./LifToken.sol");
 
-const accountGen = jsc.nat(web3.eth.accounts.length - 1);
+const knownAccountGen = jsc.nat(web3.eth.accounts.length - 1);
+const zeroAddressAccountGen = jsc.constant("zero");
```
AugustoL: this should be `0x0000000000000000000000000000000000000000` instead of zero, that is and empty address on the EVM.

test/generators.js
```diff
@@ -3,12 +3,24 @@ var jsc = require("jsverify");
...
+const accountGen = jsc.oneof([zeroAddressAccountGen, knownAccountGen]);
+
+let getAccount = (account) => {
+  if (account == "zero") {
```
AugustoL: 0x0000000000000000000000000000000000000000 here too

test/CrowdsaleGenTest.js
```diff
@@ -62,15 +62,18 @@ contract('LifCrowdsale Property-based test', function(accounts) {
...
     let shouldThrow = (publicPresaleRate == 0) ||
       (rate1 == 0) ||
       (rate2 == 0) ||
       (publicPresaleStartTimestamp >= publicPresaleEndTimestamp) ||
       (publicPresaleEndTimestamp >= startTimestamp) ||
       (startTimestamp >= end1Timestamp) ||
       (end1Timestamp >= end2Timestamp) ||
-      (setWeiLockSeconds == 0)
+      (setWeiLockSeconds == 0) ||
+      (ownerAddress == 0) ||
```
AugustoL: Here we are comparing against the address? not the address index right?

In case its against address it should be `0x0000000000000000000000000000000000000000` instad of `0`

nberger:
@AugustoL I changed the address to be `0x0000000...` now, but I get the error `could not unlock signer account`. Is it because the address does not really exist?

AugustoL:
@nberger yes its because that address doesn't exist.

nberger:
@AugustoL please review again

PR #167, opened 2017-09-06
Output testrpc memory usage to logs in travis

nberger:
testrpc is being killed in some travis runs, and it's probably
because of high memory usage. Let's output testrpc memory usage when
running in CI so we can better understand what's going on

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -2,9 +2,19 @@
 
 output=$(nc -z localhost 8545; echo $?)
 [ $output -eq "0" ] && trpc_running=true
+
+TESTRPC_REDIRECT=/dev/null
+
+# when in CI (travis), enable --mem so we can debug testrpc memory usage
+# we also need to send testrpc output to stdout instead of /dev/null so we can see it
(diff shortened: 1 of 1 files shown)


nberger:
An example of a build that failed because of testrpc being killed: https://travis-ci.org/windingtree/LifToken/jobs/272496701

Output in the log from that build is:

```
1) Contract: VestedPayment "before all" hook: prepare suite:
     Error: Could not connect to your Ethereum client. Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle.js)
      at ProviderError.ExtendableError (node_modules/truffle/build/cli.bundled.js:23642:17)
```

Enabling this flag adds the following intertwined with the tests output every few seconds (maybe just 1 second :) ):

```
{ rss: 84537344,
  heapTotal: 62824448,
  heapUsed: 43515712,
  external: 91460 }
```

kvbik:
btw this started happening on my computer too!

PR #170, opened 2017-09-07
WIP Using timestamps on MarketMaker

AugustoL:
Changing the use of blocks for timestamps on Market maker and its tests, also testing the market maker after being created by a successful crowdsale.

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -255,7 +255,7 @@ contract LifCrowdsale is Ownable, Pausable {
 
       // TODO: create the market maker with a start block that equals one month after crowdsale ends
       LifMarketMaker newMarketMaker = new LifMarketMaker(
-        address(token), block.number.add(10), 20, marketMakerPeriods, foundationWallet
+        address(token), block.timestamp.add(10), 30 days, marketMakerPeriods, foundationWallet
       );
       newMarketMaker.fund.value(mmFundBalance)();
       newMarketMaker.transferOwnership(foundationWallet);
--- a/test/MarketMaker.js
+++ b/test/MarketMaker.js
@@ -20,28 +20,21 @@ contract('marketMaker', function(accounts) {
 
   var mm;
   var token;
+  var crowdsale;
   var eventsWatcher;
 
   it("Create 24 months MM", async function() {
-    token = await help.simulateCrowdsale(100000000000, [40,30,20,10,0], accounts);
-    await help.checkToken(token, accounts, 100, [40,30,20,10,0]);
-
-    mm = await LifMarketMaker.new(
-      token.address, web3.eth.blockNumber+10, 100, 24,
-      accounts[1], {from: accounts[0]});
-
-    await mm.fund({value: web3.toWei(8, 'ether'), from: accounts[0]});
+    const mmInitialBalance = 20000000;
+    const totalTokenSupply = 100;
+    const rate = totalTokenSupply / web3.fromWei(mmInitialBalance+10000000, 'ether');
+    crowdsale = await help.simulateCrowdsale(rate, [totalTokenSupply], accounts, 1);
+    token = LifToken.at( await crowdsale.token.call());
+    mm = LifMarketMaker.at( await crowdsale.marketMaker.call());
     await mm.calculateDistributionPeriods({from: accounts[4]});
 
-    help.debug('Total Token Supply:', help.lifWei2Lif(parseFloat(await token.totalSupply.call())));
-    help.debug('MM balance:', parseInt( web3.eth.getBalance(mm.address) ));
-    help.debug('Start block', parseInt( await mm.startBlock.call() ));
-    help.debug('Blocks per period', parseInt( await mm.blocksPerPeriod.call() ));
-    help.debug('Foundation address', await mm.foundationAddr.call() );
-    help.debug('Initial Wei', parseInt( await mm.initialWei.call() ));
-    help.debug('Initial Buy Price', parseInt( await mm.initialBuyPrice.call() ));
-
-    // for (var i = 0; i < 24; i ++) { help.debug('Period', i, (await mm.marketMakerPeriods.call(i))); };
+    assert.equal(mmInitialBalance, parseInt(await web3.eth.getBalance(mm.address)));
+    assert.equal(mmInitialBalance, parseInt(await mm.initialWei.call()));
+    assert.equal(24, parseInt(await mm.totalPeriods.call()));
 
     let distributionDeltas = [
       0, 18, 99, 234, 416, 640,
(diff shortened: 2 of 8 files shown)


contracts/LifCrowdsale.sol
```diff
@@ -255,9 +255,10 @@ contract LifCrowdsale is Ownable, Pausable {
 
       // TODO: create the market maker with a start block that equals one month after crowdsale ends
       LifMarketMaker newMarketMaker = new LifMarketMaker(
-        address(token), block.number.add(10), 20, marketMakerPeriods, foundationWallet
+        address(token), block.timestamp.add(10), 30 days, marketMakerPeriods, foundationWallet
       );
       newMarketMaker.fund.value(mmFundBalance)();
+      newMarketMaker.transferOwnership(foundationWallet);
```
nberger: can we do this in a separate commit or even PR? It'd also be nice to add a test for this, for example in the `finalizeCrowdsale` test command we could assert who is the owner

test/MarketMaker.js
```diff
@@ -236,19 +223,15 @@ contract('marketMaker', function(accounts) {
...
-    const startBlock = web3.eth.blockNumber + 10;
-    const blocksPerPeriod = 15;
+    const startTimestamp = parseInt(await mm.startTimestamp.call());
+    const secondsPerPeriod = duration.days(30);
 
     const foundationWallet = accounts[9];
 
-    mm = await LifMarketMaker.new(token.address, startBlock, blocksPerPeriod, periods,
-      foundationWallet, {from: accounts[0]});
+    await mm.transferOwnership(foundationWallet, {from: accounts[0]});
```
nberger: Is this needed? Seems the crowdsale is taking care of doing this now

test/MarketMaker.js
```diff
@@ -108,81 +98,71 @@ contract('marketMaker', function(accounts) {
...
+  const tokenTotalSupply = 3000;
   let customerAddressIndex = 1;
 
   var checkScenarioProperties = async function(data, mm, customer) {
-    //help.debug("checking scenario", JSON.stringify(data));
+    help.debug("checking scenario", data.marketMakerEthBalance);
 
     assert.equal(data.marketMakerMonth, await mm.getCurrentPeriodIndex());
-    data.marketMakerEthBalance.should.be.bignumber.equal(web3.eth.getBalance(mm.address));
+    data.marketMakerEthBalance.should.be.bignumber.equal(web3.eth.getBalance(mm.address), 5);
```
nberger: why are we rounding here? Do you know where the difference comes from? Is it related to a change here?

In any case, please do it in a separate commit.

test/MarketMaker.js
```diff
@@ -191,7 +171,7 @@ contract('marketMaker', function(accounts) {
     data.marketMakerBurnedTokens.should.be.bignumber.equal(await mm.totalBurnedTokens.call());
 
     if (data.marketMakerMonth < periods) {
-      data.marketMakerBuyPrice.should.be.bignumber.equal(await mm.getBuyPrice());
+      data.marketMakerBuyPrice.should.be.bignumber.equal(await mm.getBuyPrice(), 0, BigNumber.ROUND_DOWN);
```
nberger: Same here, what's the reason for rounding here?

contracts/LifCrowdsale.sol
AugustoL: I will do this on a separate commit, and assert the owner.

test/MarketMaker.js
AugustoL: It was WIP, I was working on teh last tests of market maker, the rounding is removed now :)

test/MarketMaker.js
AugustoL: It was WIP, I was working on teh last tests of market maker, the rounding is removed now :)

test/MarketMaker.js
AugustoL: not needed, removed :)

AugustoL:
@nberger I did all the changes you mentioned, I have a separate commit with the changes on LifCrowdsale that I did to create the market maker using timestamps.

test/MarketMaker.js
```diff
@@ -20,28 +20,21 @@ contract('marketMaker', function(accounts) {
 
   var mm;
   var token;
+  var crowdsale;
```
nberger: let's avoid declaring this globally, do we need it? we can just add const every time we initialize it in the rest of the test

test/MarketMaker.js
```diff
@@ -20,28 +20,21 @@ contract('marketMaker', function(accounts) {
...
   it("Create 24 months MM", async function() {
-    token = await help.simulateCrowdsale(100000000000, [40,30,20,10,0], accounts);
-    await help.checkToken(token, accounts, 100, [40,30,20,10,0]);
-
-    mm = await LifMarketMaker.new(
-      token.address, web3.eth.blockNumber+10, 100, 24,
-      accounts[1], {from: accounts[0]});
-
-    await mm.fund({value: web3.toWei(8, 'ether'), from: accounts[0]});
+    mmInitialBalance = 20000000;
```
nberger: use `const` or `let`, the same for similar variables

test/MarketMaker.js
```diff
@@ -344,45 +328,37 @@ contract('marketMaker', function(accounts) {
     }
     assert.equal(true, thrown, "claimEth should have thrown");
 
-    try {
-      thrown = false;
-      await claimEth(0.03);
```
nberger: why was this block removed? We were asserting that claiming more than what's allowed should throw an exception, now we are not checking that

test/MarketMaker.js
```diff
@@ -108,78 +98,68 @@ contract('marketMaker', function(accounts) {
...
+    await increaseTimeTestRPC(duration.days(30)*2);
     await mm.unpause({from: accounts[0]});
-    await help.waitToBlock(web3.eth.blockNumber+blocksPerPeriod);
+    await increaseTimeTestRPC(duration.days(30));
     assert.equal(4, parseInt(await mm.getCurrentPeriodIndex()));
   });
 
   const periods = 24;
-  const tokenTotalSupply = 100;
+  const tokenTotalSupply = 3000;
```
nberger: why are we changing all the numbers in this test, starting from here?

test/VestedPayment.js
```diff
@@ -25,7 +25,8 @@ contract('VestedPayment', function(accounts) {
 
   beforeEach(async function() {
     rate = 100000000000;
-    token = await help.simulateCrowdsale(rate, [100], accounts);
+    crowdsale = await help.simulateCrowdsale(rate, [100], accounts, 1);
```
nberger: declare with const

contracts/LifCrowdsale.sol
nberger: I don't see this change in a separate commit yet, also I couldn't find the assertion. Would be great to extract it into a separate PR

test/MarketMaker.js
AugustoL: Because it was checking it twice, its checked above.

test/MarketMaker.js
AugustoL: Because the rate was less than one, thing that wont happing and in order to create the MarketMaker form the crowdsale the rate have to be higher than one.

test/MarketMaker.js
AugustoL: :+1:

test/MarketMaker.js
AugustoL: :+1:

test/VestedPayment.js
AugustoL: :+1:

test/MarketMaker.js
nberger: ok

contracts/LifCrowdsale.sol
AugustoL: https://github.com/windingtree/LifToken/pull/170/commits/6009468216dc50346998a680bcb8bd27df6ea87b here  is the commit, but you are right, the assert is missing.

test/MarketMaker.js
```diff
@@ -344,45 +328,37 @@ contract('marketMaker', function(accounts) {
...
-      await claimEth(0.03);
-    } catch(e) {
-      thrown = true;
-    }
-    assert.equal(true, thrown, "claimEth should have thrown");
-
-    // Claim 0.012 eth
-    await claimEth(0.012);
+    // Claim some ether
+    let toClaim = (await getMaxClaimableWei(state)).dividedBy(2);
```
nberger: can we use `state.marketMakerMaxClaimableWei` here, not divided by 2? We would be claiming 100% of the claimable at this point, it's a nice symmetry to previous assertion that fails with that + 1

AugustoL: :+1:

contracts/LifCrowdsale.sol
nberger: I saw that commit, but it's not only about the ownership thing. And I insist: can we move this to a separate PR? We can merge that PR before this one, it shouldn't take much time I think

AugustoL: Yes we can, I will need to edit the commit and transferOwnerhsip outside the contract on the simulateCrowdsale function. Leaving only the change on the constructor.

nberger: I think you can just start a new branch from master, add this line and the assertion in a test, and create a commit with just those changes. Once that's merged, you would merge master here, resolve the conflicts, and once this PR is finished you can rewrite the commits with a rebase

AugustoL: ok :disappointed_relieved:

nberger: Sorry if it sounds like too much work. I don't think it is, and having smaller and focused PRs helps a lot in understanding what the code is doing and why. Not only to who is reviewing the code while writing it but also to future readers of the codebase/commit history. This ownership issue is a big one that we were missing, it deserves its own PR.

PR #176, opened 2017-09-08
Add fund and finalize crowdsale composed command in property-based test

nberger:
funds the crowdsale for the amount remaining until the min cap and finalizes it. This way there's more chance the tests will also run on a finalized and funded crowdsale. Next step is to add a similar command to fund it until the market maker minimum

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -238,7 +238,7 @@ contract LifCrowdsale is Ownable, Pausable {
 
     // if the minimiun cap for the market maker is not reached transfer all funds to foundation
     // else if the min cap for the market maker is reached, create it and send the remaining funds
-    if (this.balance < foundationBalanceCapWei) {
+    if (this.balance <= foundationBalanceCapWei) {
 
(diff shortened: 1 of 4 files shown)


contracts/LifCrowdsale.sol
```diff
@@ -238,7 +238,7 @@ contract LifCrowdsale is Ownable, Pausable {
 
     // if the minimiun cap for the market maker is not reached transfer all funds to foundation
     // else if the min cap for the market maker is reached, create it and send the remaining funds
-    if (this.balance < foundationBalanceCapWei) {
+    if (this.balance <= foundationBalanceCapWei) {
```
AugustoL: If its equal it will create the MM with 0 balance

nberger: Exactly, that's what was happening, we don't want that right? We want to transfer everything up to and including `foundationBalanceCapWei` to the foundation

AugustoL: Ho yes thats the other condition, cool.

PR #177, opened 2017-09-08
Add composed command to fund crowdsale over soft cap and finalize it

nberger:
Also adds example of funding crowdsale without finalize

### DIFF
--- a/test/commands.js
+++ b/test/commands.js
@@ -316,7 +316,7 @@ let runFinalizeCrowdsaleCommand = async (command, state) => {
     let finalizeTx = await state.crowdsaleContract.finalize({from: account});
 
     let fundsRaised = state.weiRaised.div(state.weiPerUSDinTGE),
-      minimumForMarketMaker = await state.crowdsaleContract.maxFoundationCapUSD;
+      minimumForMarketMaker = await state.crowdsaleContract.maxFoundationCapUSD.call();
 
     if (crowdsaleFunded && (fundsRaised > minimumForMarketMaker)) {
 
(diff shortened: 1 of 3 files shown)


test/commands.js
```diff
@@ -316,9 +316,9 @@ let runFinalizeCrowdsaleCommand = async (command, state) => {
     let finalizeTx = await state.crowdsaleContract.finalize({from: account});
 
     let fundsRaised = state.weiRaised.div(state.weiPerUSDinTGE),
-      minimumForMarketMaker = await state.crowdsaleContract.maxFoundationCapUSD;
+      minimumForMarketMaker = await state.crowdsaleContract.maxFoundationCapUSD.call();
```
AugustoL: This should be calculated with the `state.weiPerUSDinTGE*10000000` ?

test/commands.js
```diff
@@ -564,6 +564,58 @@ let runFundCrowdsaleBelowSoftCap = async (command, state) => {
...
+    if (state.crowdsalePaused) {
+      state = await runPauseCrowdsaleCommand({pause: false, fromAccount: state.owner}, state);
+    }
+
+    // set weiPerUSDinTGE rate if needed
+    if (state.weiPerUSDinTGE == 0) {
+      state = await runSetWeiPerUSDinTGECommand({wei: 10000, fromAccount: state.owner}, state);
+    }
+
+    let softCap = await state.crowdsaleContract.maxFoundationCapUSD.call(),
```
AugustoL: same here, cant it be calculated outside the contract?

test/commands.js
```diff
@@ -564,6 +564,58 @@ let runFundCrowdsaleBelowSoftCap = async (command, state) => {
...
+
+      state = await runBuyTokensCommand(buyTokensCommand, state);
+
+      wei.should.be.bignumber.equal(state.weiRaised);
+    }
+
+    if (command.finalize) {
+      // wait for crowdsale end2Timestamp
+      if (latestTime() < state.crowdsaleData.end2Timestamp) {
+        await increaseTimeTestRPCTo(state.crowdsaleData.end2Timestamp + 1);
```
AugustoL: its necessary the +1 ?

nberger: yes, because we want it to be after the `end2Timestamp`, only after it the crowdsale can be finalized

test/commands.js
nberger: No, because it's about the `maxFoundationCapUSD` constant defined in the contract. This value is not defined in the test state AFAIK

test/commands.js
nberger: Same response :)

test/commands.js
AugustoL: https://github.com/windingtree/LifToken/blob/master/test/commands.js#L246 it is, and this value can change depending the weiPerUSD, we have the value in the state, we can use it.

test/commands.js
AugustoL: :+1:

test/commands.js
AugustoL: My bad I thought we were comparing the funds in wei, but its on usd. Its ok.

nberger: no problem, thanks!

PR #181, opened 2017-09-11
rename Market Maker to Market Validation Mechanism

AugustoL:


nberger:
There are still references to MarketMaker in README.md, LifCrowdsale.sol and commands.js

contracts/LifCrowdsale.sol
```diff
@@ -33,8 +33,8 @@ contract LifCrowdsale is Ownable, Pausable {
   // maximun balance that the foundation can have, it starts in USD
   uint256 public maxFoundationCapUSD = 10000000;
 
-  // maximun balance that the 24 month market maker can have
-  uint256 public marketMaker24PeriodsCapUSD = 40000000;
+  // maximun balance that the 24 month MVM can have
```
nberger: I'd use "Market Validation Mechanism (MVM)" at least one on each file. Like I'd you that on this comment and then just use MVM in the rest of the file. What do you think?

PR #182, opened 2017-09-11
Integrate vested tokens

AugustoL:


truffle.js
```diff
@@ -16,14 +16,15 @@ module.exports = {
       host: 'localhost',
       port: 8545,
       network_id: '*',
-      gasPrice: 100000000000
+      gas: 0x5B8D80, // 6000000 gas
+      gasPrice: 21000000000 // 21 Gwei
```
nberger: if we change the gasPrice here, we also have to change it in `helpers.js`

contracts/LifCrowdsale.sol
```diff
@@ -310,19 +358,13 @@ contract LifCrowdsale is Ownable, Pausable {
...
     if (funded()) {
 
-      // finish the minting of the token, unpause it and transfer the ownership to the foundation
-      token.finishMinting();
-      token.unpause();
-      token.transferOwnership(owner);
-
       forwardFunds();
 
+      finishDistribution();
```
nberger: I think `finishDistribution` should be part of or called from `forwardFunds`. There we know the length of the MVM so we could pass it as a parameter to `finishDistribution`, instead of having it to ask it to the MVM. That way we decouple all that part from the MVM

Another approach would be to extract small functions for each part of the `finishDistribution` and call them from `forwardFunds`. finishDistribution doesn't feel like a very good name, given we already have `finalize` and `forwardFunds`, it's not clear what belongs to where.

contracts/LifCrowdsale.sol
```diff
@@ -260,11 +267,52 @@ contract LifCrowdsale is Ownable, Pausable {
...
     }
   }
 
+  // distribute the rest of the tokens among founders and
+  function finishDistribution() internal {
+    // calculate how much tokens will the founders, foundation and advisors will receive
+    uint256 foundersTokens = token.totalSupply().mul(128).div(1000);
+    uint256 foundationTokens = token.totalSupply().mul(50).div(1000);
+    uint256 advisorsTokens = token.totalSupply().mul(72).div(1000);
+    token.mint(address(this), foundersTokens.add(foundationTokens).add(advisorsTokens));
```
nberger: F*ck, if we made the vested payment to be able to receive tokens simply via `transfer`, we wouldn't need to mint to then call to `fund`... maybe too late now but I think it would be simpler that way

AugustoL: Its almost same because we still have to call the `fund` function no matter what, if we mint the tokens after the payments are created we can avoid doing the approve. If you want I can change it nico, but I still prefer to doing this this way.

contracts/LifCrowdsale.sol
AugustoL: I thought about that but one thing is to forward the ethers raised and another its to distribute the 25% of the tokens, for me they are two separated things, and the function name its also ok for me, I mean..its what it does..it finish the distribution of the tokens. So at the end of the crowdsale it does two things: forward the ethers raised to the foundation and MVM, and it finish the distribution of the token too.

truffle.js
AugustoL: :+1:

contracts/LifCrowdsale.sol
nberger: we have to call `fund` no matter what in the current implementation of `VestedPayment`. If we made `VestedPayment` to not have a `fund` function, we wouldn't have to call it :). One way to do that is to make `VestedPayment` to calculate the percentage of funds that can be distributed on current block, and calculate that amount from the `balanceOf(address(this))`, something like that. But I understand we don't have much time to change it to work that way.

contracts/LifCrowdsale.sol
nberger: I think the name it's too abstract. And it's not even symmetric to the other stuff that is being done here. Like if it was `startDistribution` and `finishDistribution`, at least there would be some symmetry, even though I wouldn't say I like those names :). 

This is about token distribution, right? Maybe it can be called `mintAndDistributeExtraTokens`, but it shouldn't take care of the unpause/transferOwnership stuff. Also I insist that it shouldn't take the vesting length from the MVM, it should receive it as argument

contracts/LifCrowdsale.sol
AugustoL: That can work, but what happens if the contract receives more tokens in the middle of the schedule, on the next period you will be able to claim more tokens, and after the last period you will claim all of the remaining funds?

nberger: Exactly. I can't imagine why the contract would receive more tokens, except if someone wants to donate some tokens to the beneficiary of the vested payment?

contracts/LifCrowdsale.sol
AugustoL: On changing the name I agree, but not on executing the function inside forwardFunds, for me they are two separate things, and in the function we are taking in count if the MVM was created we create the vestingSchedule with the right amount of periods.

nberger: ok

contracts/LifCrowdsale.sol
AugustoL: Maybe, we didnt even started and we are receiving donations on the presale wallet, so.. if that happens from an address like an exchange, thank you very much and we took them. If we receive tokens/ethers by mistake and the sender can prove the ownership of the address we can return them after we claim it, if we want.

contracts/LifCrowdsale.sol
```diff
@@ -254,17 +263,51 @@ contract LifCrowdsale is Ownable, Pausable {
...
+      mintExtraTokens(MVMPeriods);
+
+      MVM.fund.value(mmFundBalance)();
+      MVM.transferOwnership(foundationWallet);
 
     }
   }
 
+  // distribute the rest of the tokens among founders and
+  function mintExtraTokens(uint8 foundationPaymentCliff) internal {
```
nberger: The foundation 5% reserve vesting schedule doesn't have a cliff but it starts X months after it's created. X being the MVM lifetime in 30 days periods, this vesting schedule starts on `block.timestamp + X months` and has a duration of X 30 days periods.

AugustoL:
fails on travis but no locally -.-

nberger:
> fails on travis but no locally -.-

it is failing in the coverage step in travis, perhaps it fails there because it's running all the tests... we are not running VestedPayment.js in the test step btw, I'm fixing that in #189

contracts/LifCrowdsale.sol
```diff
@@ -334,6 +343,8 @@ contract LifCrowdsale is Ownable, Pausable {
 
       foundationWallet.transfer(this.balance);
 
+      mintExtraTokens(uint256(12));
```
nberger: I think this should be 24 instead of 12. When there is no MVM, the vesting schedule for the 5% is still 2 years starting in 2 years

contracts/LifCrowdsale.sol
```diff
@@ -346,18 +357,56 @@ contract LifCrowdsale is Ownable, Pausable {
...
+      mintExtraTokens(uint256(MVMPeriods));
+
+      MVM.fund.value(mmFundBalance)();
+      MVM.transferOwnership(foundationWallet);
 
     }
   }
 
   /**
+     @dev Internal. Distribute the rest of the tokens among founders and advisors
```
nberger: I'd explain what's the exact distribution, something like: "Distribute extra tokens among founders, team and the foundation long-term reserve. Founders receive 12.8% of tokens in a 4y (1y cliff) vesting schedule. Foundation long-term reserve receives 5% of tokens in a vesting schedule with the same duration as the MVM that starts when the MVM ends. An extra 7.2% is transferred to the foundation to be distributed among advisors and future hires"

contracts/LifCrowdsale.sol
```diff
@@ -346,18 +357,56 @@ contract LifCrowdsale is Ownable, Pausable {
...
+   */
+  function mintExtraTokens(uint256 foundationMonthsStart) internal {
+    // calculate how much tokens will the founders, foundation and advisors will receive
+    uint256 foundersTokens = token.totalSupply().mul(128).div(1000);
+    uint256 foundationTokens = token.totalSupply().mul(50).div(1000);
+    uint256 advisorsTokens = token.totalSupply().mul(72).div(1000);
```
nberger: It's not only advisors, it's also future hires. I'd call it `teamTokens`

contracts/LifCrowdsale.sol
```diff
@@ -346,18 +357,56 @@ contract LifCrowdsale is Ownable, Pausable {
...
+
+    // create the vested payment schedule for the founders
+    foundersVestedPayment = new VestedPayment(
+      block.timestamp, 30 days, 48, 12, foundersTokens, token
+    );
+    token.mint(foundersVestedPayment, foundersTokens);
+    foundersVestedPayment.transferOwnership(foundationWallet);
+
+    // create the vested payment schedule for the foundation
+    uint256 foundationPaymentStart = foundationMonthsStart.mul(604800);
```
nberger: 604800 is a week, I think we are talking about months here. Maybe it's better to use `foundationMonthsStart.mul(1 month)` (or `30 days` instead of `1 month`)

contracts/LifCrowdsale.sol
```diff
@@ -346,18 +357,56 @@ contract LifCrowdsale is Ownable, Pausable {
...
+
+    // finish the minting of the token and unpause it
+    token.finishMinting();
+    token.unpause();
+
+    // transfer the ownership of the token to the foundation
+    token.transferOwnership(owner);
+  }
+
+  // @return true if the transaction can buy tokens on TGE
```
nberger: remove this line

contracts/LifCrowdsale.sol
```diff
@@ -346,18 +357,56 @@ contract LifCrowdsale is Ownable, Pausable {
...
+      block.timestamp.add(foundationPaymentStart), 30 days,
+      foundationMonthsStart, 0, foundationTokens, token
+    );
+    token.mint(foundationVestedPayment, foundationTokens);
+    foundationVestedPayment.transferOwnership(foundationWallet);
+
+    // transfer the token for advisors and future employees to the foundation
+    token.mint(foundationWallet, advisorsTokens);
+
+    // finish the minting of the token and unpause it
```
nberger: maybe it's better to do the `finishMinting`, `unPause` and `transferOwnership` in `finalize`, after the call to `forwardFunds`, what do you think?

test/MarketMaker.js
```diff
@@ -174,7 +174,7 @@ contract('Market validation Mechanism', function(accounts) {
 
     assert.equal(data.MVMMonth >= periods, await mm.isFinished());
 
-    data.ethBalances[customerAddressIndex].should.be.bignumber.equal(web3.eth.getBalance(customer));
+    // data.ethBalances[customerAddressIndex].should.be.bignumber.equal(web3.eth.getBalance(customer));
```
nberger: is this line still failing, or can we uncomment it already?

contracts/LifCrowdsale.sol
AugustoL: we cant do the finishMinting before, we need to mint the extra tokens inside forwardFunds.

test/MarketMaker.js
AugustoL: :+1:

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol
nberger: I agree we can't do it before `forwardFunds`, that's why I'm saying to do it after the call to `forwardFunds` ;)

test/MarketMaker.js
AugustoL: I think we are doing something wrong now on the eth balance calculation on the test once the MVM buy the tokens back, the test fail when now it takes the 25% extra tokens in count.

nberger:
I approve as soon as https://github.com/AugustoL/LifToken/pull/5 gets merged :)

Issue #184, opened 2017-09-11
transferData: fail when `call` returns false


nberger:
`transferData` uses `call` to execute a contract call on transfer. But the `call` return value is being ignored, so if the `call` throws, the `transferData` does not throw. I think that should be avoided: if the `call` fails, the `transferData` should also fail. And this is connected to #183: "fail" currently means that it should throw, but if we change it to return bool instead then fail would mean to return false. In that case we could just return the result from the `call`.

For context, here's the documentation of `call`: https://solidity.readthedocs.io/en/latest/types.html#members-of-addresses

/cc @AugustoL

AugustoL:
thx will look at it tomorrow.

AugustoL:
fixed in #191

PR #191, opened 2017-09-13
Add return value on new transfer methods in token.

### LINKED ISSUE #183: make transferData and related functions to return bool, as the analog methods in ERC20

AugustoL:
fix #183 #184

### DIFF
--- a/contracts/LifToken.sol
+++ b/contracts/LifToken.sol
@@ -44,6 +44,8 @@ contract LifToken is MintableToken, Pausable {
      @param value The amount of tokens to be spent.
      @param data ABI-encoded contract call. For example generated using web3's
      getData method
+
+     @return true if the call function was executed successfully
    */
   function approveData(address spender, uint value, bytes data) whenNotPaused returns (bool) {
 
@@ -68,6 +70,8 @@ contract LifToken is MintableToken, Pausable {
      @param value uint256 the amout of tokens to be transfered
      @param data ABI-encoded contract call. For example generated using web3's
      getData method
+
+     @return true if the call function was executed successfully
    */
   function transferData(address to, uint value, bytes data) whenNotPaused returns (bool) {
 
@@ -97,6 +101,8 @@ contract LifToken is MintableToken, Pausable {
      @param value The amout of tokens to be transferred
      @param data ABI-encoded contract call. For example generated using web3's
      getData method
+
+     @return true if the call function was executed successfully
    */
   function transferDataFrom(address from, address to, uint value, bytes data) whenNotPaused returns (bool) {
 
--- a/contracts/LifToken.sol
+++ b/contracts/LifToken.sol
@@ -45,14 +45,18 @@ contract LifToken is MintableToken, Pausable {
      @param data ABI-encoded contract call. For example generated using web3's
      getData method
    */
-  function approveData(address spender, uint value, bytes data) {
+  function approveData(address spender, uint value, bytes data) whenNotPaused returns (bool) {
 
     require(spender != address(this));
 
     allowed[tx.origin][spender] = value;
 
-    if (spender.call(data))
+    if (spender.call(data)) {
       ApprovalData(tx.origin, spender, value, data);
+      return true;
+    } else {
+      return false;
+    }
 
   }
 
@@ -65,7 +69,7 @@ contract LifToken is MintableToken, Pausable {
      @param data ABI-encoded contract call. For example generated using web3's
      getData method
    */
-  function transferData(address to, uint value, bytes data) {
+  function transferData(address to, uint value, bytes data) whenNotPaused returns (bool) {
 
     require(to != address(this));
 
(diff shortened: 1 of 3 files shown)


test/LifToken.js
```diff
@@ -132,6 +132,83 @@ contract('LifToken', function(accounts) {
...
+
+    try {
+      transaction = await token.approveData(
+        message.contract.address, help.lif2LifWei(10), data,
+        {from: accounts[1]}
+      );
+    } catch (error) {
+      if (!help.isInvalidOpcodeEx(error)) throw error;
+
+      let decodedEvents = help.abiDecoder.decodeLogs(transaction.receipt.logs);
```
nberger: probably this code is not executing: `approveData` is not supposed to throw now, it's supposed to return false. If it were throwing then `transaction` would not have been assigned and this would fail because transaction being undefined in `transaction.receipt.logs`, right?

So I think instead of the try/catch we should just assert that the return value from `approveData` equals false

test/LifToken.js
```diff
@@ -132,6 +132,83 @@ contract('LifToken', function(accounts) {
...
+    let data = message.contract.fail.getData();
+    let transaction;
+
+    try {
+      transaction = await token.transferData(
+        message.contract.address, help.lif2LifWei(10), data,
+        {from: accounts[1]}
+      );
+    } catch (error) {
+      if (!help.isInvalidOpcodeEx(error)) throw error;
```
nberger: same here as for `approveData`, this code is probably not running

test/LifToken.js
```diff
@@ -132,6 +132,83 @@ contract('LifToken', function(accounts) {
...
+
+    await token.approve(accounts[1], help.lif2LifWei(10), {from: accounts[2]});
+
+    try {
+      transaction = await token.transferDataFrom(
+        accounts[2], message.contract.address, help.lif2LifWei(10), data,
+        {from: accounts[1]}
+      );
+    } catch (error) {
+      if (!help.isInvalidOpcodeEx(error)) throw error;
```
nberger: and same here, this code doesn't run

test/LifToken.js
AugustoL: Hm yes, right now for tetsing in throwing inside the fail() function called on the approve, but it should return false instead of throw.
We cant access the return value when sending transaction, that works only for constant functions when are called and not executed via transactions.
Im going to return false on the fail() function and test that there is not events on the receipt.logs.

nberger: there's nothing wrong in throwing inside of `fail`, that's what we need: `call` returns false if the function has thrown, true otherwise

nberger: I see it's difficult to check whether it returned false or true... we would need another contract to issue the `approveData` and in that contract we could have some logic that depending on that return value would update some state in that contract that we can check later in the test.

In the end these `approveData`, `transferData`, etc are intended to be called from other contracts I think so it makes sense to test it that way, or am I missing something?

AugustoL: So its ok, we know if it returned false or true depending on the events triggered on the transaction, we have tests where the events are triggered and we have tests where they are not, and that its because false was returned form the call. On the other commit I added the tests for that,

AugustoL: On this commit, sorry, the one with the tests.

nberger: ok, makes sense, but the try/catch is misleading, we should remove it, that code never executes. If we think the catch executes, then we should add a throw in the try block after the `approveData` call

test/LifToken.js
```diff
@@ -132,6 +132,83 @@ contract('LifToken', function(accounts) {
     await help.checkToken(token, accounts, 125, [40,30,20,10,0]);
   });
 
+  it('should fail when calling a function that fails inside approveData', async function() {
```
nberger: I think it would be nice to explain how it fails here, like `should fail and not trigger an ApproveData event...`

test/LifToken.js
nberger: so there's one more piece missing here: as the catch is not running, the `assert.equal(0, decodedEvents.length);` is not running either, we have to keep that code but without the try/catch, makes sense?

contracts/LifToken.sol
```diff
@@ -45,14 +45,18 @@ contract LifToken is MintableToken, Pausable {
      @param data ABI-encoded contract call. For example generated using web3's
      getData method
    */
-  function approveData(address spender, uint value, bytes data) {
+  function approveData(address spender, uint value, bytes data) whenNotPaused returns (bool) {
```
nberger: can you please add the `@return true ...` to the doxygen comment?

AugustoL: Yes! What do you think of : 
`@return true if the call function was executed successfully`
On the three transfer data functions?

nberger: what about `@return true if the call with data did not throw`?

PR #194, opened 2017-09-13
Remove public presale

AugustoL:


nberger:
This comment in LifCrowdsale might need update, I think it refers to the public presale, we should remove that part:

>  token sale, with a discounted rate during the 1st week and discounted rates for contributions during the presale period and a Market Validation Mechanism that will receive the funds over the USD 10M

Also this one (remove presale):

> // Seconds before the Presale and the TGE since when the corresponding USD to

Also (remove public):

> // Amount of wei received as presale payments (both private and public)

Also remove buyPresaleTokens:

>    @dev Fallback function, payable. Calls `buyTokens` or `buyPresaleTokens`

PR #196, opened 2017-09-13
Fix for linter: remove unused var in test

nberger:


nberger:
Closing, fixing it in #195

nberger:
Let's better merge it so #198 finally fixes the build :)

PR #200, opened 2017-09-14
Private presale custom rates

AugustoL:


contracts/LifCrowdsale.sol
```diff
@@ -238,13 +231,17 @@ contract LifCrowdsale is Ownable, Pausable {
 
      @param beneficiary Address to which Lif will be sent
      @param weiSent Amount of wei contributed
+     @param rate of tokens per ether at the moment of the contribution
```
nberger: I'd change it to:

`@param rate Lif per ether rate at the moment of the contribution`

(changed tokens to Lif, and also note that rate is the param name, and the rest is the description of the param, it's not supposed to be a regular phrase as a whole but first the param, then the description)

contracts/LifCrowdsale.sol
```diff
@@ -238,13 +231,17 @@ contract LifCrowdsale is Ownable, Pausable {
...
    */
-  function addPrivatePresaleTokens(address beneficiary, uint256 weiSent) onlyOwner {
+  function addPrivatePresaleTokens(
+    address beneficiary, uint256 weiSent, uint256 rate
+  ) onlyOwner {
     require(block.timestamp < startTimestamp);
     require(beneficiary != address(0));
     require(weiSent > 0);
+    require(rate > rate1);
```
nberger: I'd add a comment explaining this. It'd be something like "validate that rate is higher than TGE rate"

contracts/LifCrowdsale.sol
```diff
@@ -238,13 +231,17 @@ contract LifCrowdsale is Ownable, Pausable {
...
 
-    uint256 tokens = weiSent.mul(privatePresaleRate);
+    uint256 tokens = weiSent.mul(rate);
```
nberger: Would adding an event here give an easier way to see the details of the presale contributions? Should we add one?

contracts/LifCrowdsale.sol

contracts/LifCrowdsale.sol
AugustoL: we can add one, yes. It will be easier.

PR #215, opened 2017-09-18
More MVM property-based tests

### LINKED ISSUE #156: Add property-based tests for the Market Validation Mechanism (Market Maker)

nberger:
Adds a specific example that was failing, refactors the MVM tests by adding its commands as property-based test commands in `commands.js`

Fixes #156

### DIFF
--- a/test/commands.js
+++ b/test/commands.js
@@ -456,31 +456,11 @@ async function runTransferFromCommand(command, state) {
   return state;
 }
 
-
 //
-// Market Maker commands
+// Composed commands
 //
-
-let getMvmMaxClaimableWei = function(state) {
-  if (state.MVMMonth >= state.MVMPeriods) {
-    help.debug('calculating maxClaimableEth with', state.MVMStartingBalance,
-      state.MVMClaimedWei,
-      state.returnedWeiForBurnedTokens);
-    return state.MVMStartingBalance.
-      minus(state.MVMClaimedWei).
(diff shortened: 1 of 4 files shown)


nberger:
@AugustoL this is mostly done, I just want to add the MVMWaitToMonth command that simulates waiting for a specific month in the MVM lifetime. Oh, we should also add an MVMPause command, that's an easy one and once that's done we have everything covered I think

test/commands.js
```diff
@@ -633,7 +662,7 @@ async function runMVMSendTokensCommand(command, state) {
       state.MVMBurnedTokens = state.MVMBurnedTokens.plus(lifWei);
       state.returnedWeiForBurnedTokens = state.returnedWeiForBurnedTokens.plus(tokensCost);
       state.balances[command.from] = getBalance(state, command.from).minus(lifWei);
-      state.MVMMaxClaimableWei = getMMMaxClaimableWei(state);
+      state.MVMMaxClaimableWei = getMvmMaxClaimableWei(state);
```
AugustoL: here it shouldnt be `getMVMMaxClaimableWei` to follow the use of MVM on uppercase on every variable ?

nberger: yeah, I changed it because it was too many `M`s together :), I thought this way had better readability. Not sure whether should we favor consistency or readability :)

what do you think?

AugustoL: I think it still readable, I would prefer to follow the consistency in this case.

PR #223, opened 2017-09-21
Fix addPrivatePresalePayment: add to totalPresaleWei var

nberger:


nberger:
The build failed in the coverage step (after running the tests the first time fine) on a test that is fixed in the first commit in #215 so I'll merge this PR

Issue #178, opened 2017-09-08
Parallelize tests and coverage in travis


nberger:
We should do something similar to https://github.com/OpenZeppelin/zeppelin-solidity/pull/369/files so test and coverage run in parallel

nberger:
The complication here is that we are already [running the tests in 2 steps](https://github.com/windingtree/LifToken/blob/7f58c1330650a011ef0d67584ddf0142b695811e/.travis.yml#L12-L13) so even though it's doable, adding the parallelization would complicate the build quite a bit and I'm not sure how worth it is. Maybe it's better to keep it simple even when it takes longer to run and sometimes it fails because of failures in the coverage step (which might be a good thing).

PR #229, opened 2017-09-21
Upgrade solidity coverage

nberger:


nberger:
I'm not merging this yet because it keeps failing in travis with:

```
1) Contract: Market validation Mechanism should go through scenario with some claims and sells on the Market Maker:
     AssertionError: expected '9.999999999999999999999999999999999999999999999999999999999715998399999849121107e+78' to equal '9.999999999999999999999999999999999999999999999999999999999715998399999849217441e+78'
 ```

But when I run it locally it fails with a different error, the dreaded `NotFoundError: Key not found in database` from testrpc after which testrpc seems to die and that's it...

@AugustoL can you try running `npm run coverage` (or `yarn coverage`, I'm actually running with yarn) after doing `yarn install` on this branch and see what happens?

nberger:
Now the build is green even though the coverage part is failing (the build doesn't fail because we made the coverage task to not fail the travis build in #178). We still have to fix the issue before merging this.

cgewecke:
@nberger Is the test that's failing a test of someone's eth balance after sending to a fallback?

nberger:
@cgewecke yes, it's a test of someone's eth balances, even though there's no fallback involved AFAICT. It's on a test that runs the following transaction: https://github.com/windingtree/LifToken/blob/9164e168837216d72ece26cab52bade06fd7d8ce/contracts/LifMarketValidationMechanism.sol#L234-L248. The assertion that fails is this one: https://github.com/windingtree/LifToken/blob/9164e168837216d72ece26cab52bade06fd7d8ce/test/MarketValidationMechanism.js#L219

cgewecke:
Oh I think this is a bad bug in solidity-coverage that was discovered last week. I'm pushing a fix rn. Maybe upgrading one more step will make the problem go away :) What about the `invalid opcode` failure in the crowdsale test?

cgewecke:
@nberger Maybe try bumping to `0.2.5`?

nberger:
Yes, I'll try that and come back with the result. Thanks @cgewecke!

nberger:
@cgewecke yes! it seems the test runs fine in 0.2.5, at least it does in my local env. Running in travis now 🤞

nberger:
I'm going to merge this, thanks again @cgewecke!!

> What about the invalid opcode failure in the crowdsale test?

Which test are you referring to?

cgewecke:
Oh I guess it's gone! I thought I saw that in the old travis run. Good
news, and sorry for the headaches with this.

PR #230, opened 2017-09-21
Add pause & unpause MVM test commands

### LINKED ISSUE #222: Add test for pause/unpause MVM
Related to #156

nberger:
Also adds Pause & Unpause events in MVM

Fixes #222

### DIFF
--- a/test/commands.js
+++ b/test/commands.js
@@ -649,8 +649,10 @@ let isMVMFinished = (state) => state && false;
 async function runMVMClaimEthCommand(command, state) {
   if (state.MVM !== undefined) {
     let weiToClaim = web3.toWei(command.eth),
-      hasZeroAddress = false,
-      shouldThrow = (weiToClaim > getMVMMaxClaimableWei(state));
+      hasZeroAddress = false;
+
+    let shouldThrow = (weiToClaim > getMVMMaxClaimableWei(state)) ||
+      state.MVMPaused;
 
     try {
       help.debug('Claiming ', weiToClaim.toString(), 'wei (', command.eth.toString(), 'eth)');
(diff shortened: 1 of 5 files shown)


test/commands.js
```diff
@@ -766,6 +768,28 @@ async function runMVMWaitForMonthCommand(command, state) {
...
+      shouldThrow = (state.MVMPaused == command.pause) || (command.fromAccount != state.owner);
+
+    try {
+      if (command.pause) {
+        await state.MVM.pause({from: fromAccount});
+      } else {
+        await state.MVM.unpause({from: fromAccount});
+      }
+
+      state.MVMPaused = command.pause;
```
AugustoL: It will be useful to save the timestamp when it was paused, right? Maybe save that instead on a boolean, if it was paused it will be higher than 0. And we can compare the total time paused too, that is used on `getCurrentIndex()` if you try to get the current index after it was paused are you not receiving any error comparing to the expected one?

nberger: Yes, you are right, it's important to keep track of that in the tests too, to verify it works well. I'll add it :+1:

nberger:
@AugustoL: I just pushed the changes to track the paused seconds in the prop-based test. Added a few other options as I was there. How does it look now?

AugustoL:
Much better, but I think we still have to take in count the totalPausedSeconds when we get the current period index in the tests. We take in count that in the contract.

nberger:
Ok, I see your point, I agree that we can test that better. The month is being stored in `state.MVMMonth`, and it's only updated from the `MVMWaitForMonth` command. First thing would be to take the secondsPaused into account in that command. Second thing would be to also update the `MVMMonth` from the `waitTime` command. That would be it, right? Or anything else we are missing? Maybe also add a specific example in `test/MarketValidationMechanism.js` if we have time?

AugustoL:
I think we can only change the the part where the MVMMonth is calculated, right now in MVMWaitForMonth, or we shoudl need to add it on another place?

PR #231, opened 2017-09-22
README: fix typo and rephrase intro paragraph

nberger:


nberger:
Being honest: this PR is made mostly to be able to merge it fast so as to trigger a build in master to have the build and coverage badge updated (it hasn't updated for a while, coverage is 80% there but should be around 87%, I enabled travis builds for all pushes so I think the badge should update once this is merged)

PR #235, opened 2017-09-25
Fix burn token underfunded crowdsale test

nberger:


nberger:
I'll merge even when one test is failing, because it only fails in Travis (I can't reproduce in my local environment) and anyways this change is not causing that failure

PR #242, opened 2017-09-26
Fix MVM creation: add missing calculate periods call

nberger:


nberger:
@AugustoL I saw your comment about moving the SentTokens event to a separate PR too late, sorry! :)

PR #254, opened 2017-09-27
Add ClaimedEth event on MVM claimEth calls

### LINKED ISSUE #253: Add an event for when the foundation claims ether from the MVM
We have an event for `sendTokens` but we are missing an event for `claimEth`

Thanks to CoinFabrik for reporting this issue as part of their audit

nberger:
Fixes #253

### DIFF
--- a/contracts/LifMarketValidationMechanism.sol
+++ b/contracts/LifMarketValidationMechanism.sol
@@ -62,6 +62,7 @@ contract LifMarketValidationMechanism is Ownable {
   event Pause();
   event Unpause(uint256 pausedSeconds);
 
+  event ClaimedEth(uint256 claimedWei);
   event SentTokens(address indexed sender, uint256 price, uint256 tokens, uint256 returnedWei);
 
   modifier whenNotPaused(){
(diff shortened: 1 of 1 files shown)


contracts/LifMarketValidationMechanism.sol
```diff
@@ -62,6 +62,9 @@ contract LifMarketValidationMechanism is Ownable {
   event Pause();
   event Unpause(uint256 pausedSeconds);
 
+  event ClaimedEth(uint256 claimedWei);
```
AugustoL: it shouldnt be `claimedWei`? Its in wei units :)

nberger: I wanted to keep it consistent with the corresponding method name `claimEth` which also deals with wei amounts but it has eth in its name :)

Should we change the method name too? Maybe in a subsequent PR?

AugustoL: Yes I think we should, it claims it in Wei unit. So the next PR can be change Eth to Wei in naming.

nberger: Cool, I'll change both in a further PR, thanks!

PR #256, opened 2017-09-27
Upgrade to solidity 0.4.17

### LINKED ISSUE #244: Upgrade to solidity compiler 0.4.15
Current version being used (0.4.13) is known to have two important vulnerabilities:

* DelegateCallReturnValue
The return value of the low-level .delegatecall() function is
taken from a position in memory, where the call data or the return data resides. This
value is interpreted as a boolean and put onto the stack. This means if the called
function returns at least 32 zero bytes, .delegatecall() returns false even if the call
was successuful."

* ECRecoverMalformedInput
The ecrecover precompile does not properly signal
failure for malformed input (especially in the 'v' argument) and thus the Solidity
function can return data that was previously present in the return area in memory."

nberger:
Fixes #244

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "ethereumjs-abi": "^0.6.4",
     "jsverify": "^0.8.2",
     "lodash": "^4.17.4",
-    "solc": "^0.4.13",
+    "solc": "0.4.17",
     "truffle-hdwallet-provider": "0.0.3",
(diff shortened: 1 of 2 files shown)


nberger:
Closing in favor of #257: let's also upgrade truffle and update solc to 0.4.15 only as that's what truffle is using

PR #266, opened 2017-09-28
Add missing Transfer and Approval events

### LINKED ISSUE #248: Missing Transfer events
Add Transfer events for `transferData` and `transferDataFrom` calls in LifToken. Consider also adding a Transfer event to address 0x0 in `burn` so wallets and block explorers (like etherscan) can show the right balances

nberger:
Added the events in the *Data methods, and in burn

Fixes #248

### DIFF
--- a/contracts/LifToken.sol
+++ b/contracts/LifToken.sol
@@ -55,6 +55,7 @@ contract LifToken is MintableToken, Pausable {
 
     if (spender.call(data)) {
       ApprovalData(tx.origin, spender, value, data);
+      Approval(tx.origin, spender, value);
       return true;
     } else {
       return false;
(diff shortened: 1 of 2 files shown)


nberger:
Don't we want to have an event that includes the data param? Using the original events only has the downside that they don't include that value

AugustoL:
Yes, but the data parameter can be extracted from the transaction data, I think that will be better because if not we will log more information than the one we need. Ltes just use the ERC20 token events.

nberger:
Makes sense, I'll change it that way

nberger:
@AugustoL: just made the changes, how does it look now?

AugustoL:
Great, thanks @nberger

PR #274, opened 2017-10-03
Assert on finalize that gasUsed should be less than limit

### LINKED ISSUE #247: Too complex finalize function in LifCrowdsale
The finalize operation creates several contracts to manage Founders, Foundation and Developers tokens, and depending on the total raised it can create a LifMarketValidationMechanism contract.

We should try and make this function simpler, to avoid hitting the gas limit in the mainnet, and also to not force the caller of finalize to pay the gas for all the work needed to finalize the crowdsale

nberger:
LifCrowdsale `finalize` is a quite expensive transaction because it creates a few contracts (MVM and VestedPayments). Let's assert that it uses less gas than the limit in mainnet

Fixes #247

### DIFF
--- a/test/commands.js
+++ b/test/commands.js
@@ -247,6 +247,8 @@ async function runFinalizeCrowdsaleCommand(command, state) {
     help.debug('finishing crowdsale on block', nextTimestamp, ', from address:', gen.getAccount(command.fromAccount), ', funded:', crowdsaleFunded);
 
     let tx = await state.crowdsaleContract.finalize({from: account});
+    help.debug('gas used in finalize:', tx.receipt.gasUsed);
+    assert(tx.receipt.gasUsed < 6700000, 'gas used in finalize should be less than gas limit in mainnet');
 
     let fundsRaised = state.weiRaised.div(state.weiPerUSDinTGE),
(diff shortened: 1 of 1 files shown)


nberger:
@AugustoL: I tried to simplify the `finalize`, `forwardFunds` and `mintExtraTokens` methods in `LifCrowdsale` but it's not super easy, we'd need to do some changes: if we want to delay creating the VestedPayments to a further transaction we'd need to avoid calling `finishMinting` in the finalize, but that's kind of risky, we want to finish minting before unpausing the token. We'd also need to save some of the calculations done in `forwardFunds` in some state var to be used later on this other transaction (amounts for vestedPayments, periods for the MVM, etc...)... So I'd avoid doing those changes and just make sure that we are fine with the gas used. I'm adding the assertion here and from the debug statements it seems we are fine, here's the 3 different values I get when running the tests locally:

```
gas used in finalize: 2989489
gas used in finalize: 1133634
gas used in finalize: 3004489
```

What do you think? Can we close #274 with this?

AugustoL:
Yes, lets close it for now, we will know if its too much once we test it in main net.

PR #279, opened 2017-10-05
Check eth balances in property-based test

### LINKED ISSUE #278: Check eth balances of all accounts in property-based test
It'd be nice to check the eth balances of all the accounts at the end of the tests. For this to work, we need to keep track of all the eth movements, starting from the initial balances when the test starts.

What makes this difficult is the gas cost: each transaction involves some gas spending. In successful transactions we can easily take the spent gas from the tx receipt, as we are already doing in the [MVM sendTokens test](https://github.com/windingtree/LifToken/blob/dd4b2443fa206f57992a24a0bbd410cb221c782d/test/commands.js#L724-L726). The problem is with failed transactions: in those cases an exception is thrown and there's no easy way to know how much gas was spent (web3.js doesn't give us the txId or the tx receipt, even though testrpc logs those transaction details to the console).

I started working on this in the `check-eth-balances` branch, will open a [WIP] PR soon

nberger:
Fixes #278

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -7,5 +7,5 @@ if [ "$SOLIDITY_COVERAGE" = true ]; then
 else
   yarn lint
   WT_DEBUG=true yarn test test/LifToken.js test/Crowdsale.js test/MarketValidationMechanism.js test/VestedPayment.js
-  WT_DEBUG=true GEN_TESTS_QTY=40 yarn test test/CrowdsaleGenTest.js
+  WT_DEBUG=true GEN_TESTS_TIMEOUT=400 GEN_TESTS_QTY=40 yarn test test/CrowdsaleGenTest.js
 fi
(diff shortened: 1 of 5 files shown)


AugustoL:
I think it will be easier to have just a function that receives the tx object and and it decrease or increase the balance on the addresses depending on the sender and receiver of the tx, the sender or receiver can be eth accounts or contract addresses. You dont think it would be a better solution @nberger ? like a function `processTxBalances(tx)`

nberger:
@augustol: I agree on that would make it easier, but the downside is that we end relying on the final values read from the blockchain instead of the values that we expect should be the result of the txs. We want to verify that the values in the blockchain are the same as we expect, to do that we cannot read the values from the blockchain except at the end to verify that they are the same as we tracked command after command. That's how I see it. Or maybe I'm missing something?

AugustoL:
Yes, I agree, we should tests it with the values that we expect.

PR #280, opened 2017-10-06
Refactor *Data methods in LifToken to use super methods

nberger:
Calls super.approve, super.transfer & super.transferFrom from the corresponding
*Data methods (approveData, transferData & transferDataFrom)

### DIFF
--- a/contracts/LifToken.sol
+++ b/contracts/LifToken.sol
@@ -53,8 +53,8 @@ contract LifToken is MintableToken, Pausable {
 
     allowed[msg.sender][spender] = value;
 
-    assert(spender.call(data));
     Approval(msg.sender, spender, value);
+    assert(spender.call(data));
 
     return true;
   }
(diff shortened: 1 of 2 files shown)


contracts/LifToken.sol
```diff
@@ -48,18 +48,11 @@ contract LifToken is MintableToken, Pausable {
...
-
-    if (spender.call(data)) {
-      Approval(tx.origin, spender, value);
-      return true;
-    } else {
-      return false;
-    }
-
+    super.approve(spender, value);
+    assert(spender.call(data));
```
AugustoL: I think here we should use require instead of assert.

nberger: Agreed, I just changed it

contracts/LifToken.sol
```diff
@@ -105,22 +87,11 @@ contract LifToken is MintableToken, Pausable {
...
-
-    if (to.call(data)) {
-      Transfer(tx.origin, to, value);
-      return true;
-    } else {
-      return false;
-    }
-
+    super.transferFrom(from, to, value);
+    require(to.call(data));
```
AugustoL: I think it should be first the require to save some gas, because if it fails later the gas used by transfer will be consumed.

nberger: Agreed, just changed it in d818a37

PR #282, opened 2017-10-06
Do not run property-based tests in Coverage

nberger:
We should aim at a high coverage even without the property-based tests


nberger:
Coverage went up to 93.55 (around +1%) so we should merge this

PR #289, opened 2017-10-10
And more tests and increase coverage in LifCrowdsale

nberger:


nberger:
Coverage is > 99% now. Can't make it higher without removing/changing code: only 2 branches are missing and it's in validations that are unreachable because of previous validations. I think it's fine as it is :)

PR #298, opened 2017-10-12
[WIP] Upgrade to Truffle 4.0.0

nberger:
Upgrades to solc 0.4.17 and truffle 4.0.0-beta.2

But perhaps we should wait until the final truffle 4.0.0 is released before merging this

@AugustoL, @cgewecke: thoughts?

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.15;
+pragma solidity ^0.4.17;
 
 import "zeppelin-solidity/contracts/ownership/Ownable.sol";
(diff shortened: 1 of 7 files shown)


cgewecke:
@nberger Yeah! It works. Looks good.

There are a couple bugs at solidity-coverage we have to fix (today?) so that `pure` ,`view` and `pragma experimental` don't cause problems.  But they only affect files imported from outside the contracts folder (pure, view) and yarn installations (pragma experimental).

AugustoL:
Lets continue using solc 0.4.15, we dont want to upgrade and change the contracts again.

nberger:
Ok, closing now then, thanks!

PR #301, opened 2017-11-15
Use BurnableToken and PausableToken

cwhinfrey:
* Make LifToken burnable, remove copied code
* Use PausableToken instead of Pausable


AugustoL:
Thanks @cwhinfrey !

PR #305, opened 2017-12-21
Add Lif Channels

AugustoL:
The LifChannels contract is an simple implementation of state channels for ERC20 tokens. This contract allows the token holders to mediate and decide about how much the final value of a transaction is going to be.
The channel is opened by a token holder that previously approved the transfer of tokens to the LifChannels contract, once the channel is opened the tokens are transfered to the contract and the mediation of the funds start.
While the channel is opened the two parties can exchange signatures proving the allowance and claim of the funds to be transfered, in this case the sender will share signatures where he allows the receiver to claim certain amount of tokens and the receiver will share signatures where he is willing to claim that amount.
With those signatures the receiver and sender can reach a mutual agreement on how much the final value of the transfer is going to be and always have the time to prevent the another party to close the channel with the wrong final value.

### DIFF
--- a/contracts/LifChannels.sol
+++ b/contracts/LifChannels.sol
@@ -5,7 +5,7 @@ import "zeppelin-solidity/contracts/math/SafeMath.sol";
 import "zeppelin-solidity/contracts/ECRecovery.sol";
 
 /**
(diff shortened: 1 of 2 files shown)


kvbik:
WOW!!!

PR #306, opened 2018-01-04
Add return purchase function in crowdsale

AugustoL:
We need to add a function to return the purchase of tokens from contributors in the crowdsale to reject contributions from unverified users. The return of purchases will be executed after the crowdsale ends and before the finalization of the crowdsale.

This changes add that function and change the tests to test it correctly.

### DIFF
--- a/contracts/LifCrowdsale.sol
+++ b/contracts/LifCrowdsale.sol
@@ -417,9 +417,8 @@ contract LifCrowdsale is Ownable, Pausable {
 
       forwardFunds();
 
-      // finish the minting of the token and unpause it
+      // finish the minting of the token
       token.finishMinting();
-      token.unpause();
 
       // transfer the ownership of the token to the foundation
       token.transferOwnership(owner);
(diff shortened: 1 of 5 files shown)


contracts/LifCrowdsale.sol
```diff
@@ -387,6 +387,22 @@ contract LifCrowdsale is Ownable, Pausable {
...
+   */
+  function returnPurchase(address contributor)
+    public hasEnded onlyOwner
+  {
+    require(!isFinalized);
+
+    uint256 toReturn = purchases[contributor];
+    assert(toReturn > 0);
+
+    purchases[msg.sender] = 0;
```
adklempner: ❗️ this should be `purchases[contributor] = 0;`

contracts/LifCrowdsale.sol
```diff
@@ -387,6 +387,22 @@ contract LifCrowdsale is Ownable, Pausable {
...
+
+    msg.sender.transfer(toReturn);
```
adklempner: this should be `contributor.transfer`

AugustoL: good catch!

PR #309, opened 2018-01-18
Rename SmartToken for ERC827

AugustoL:
Replace SmartToken for ERC827 token standard in contracts and test folder.


contracts/ERC827Token.sol
```diff
@@ -29,7 +29,7 @@ contract SmartToken is StandardToken {
 
      @return true if the call function was executed successfully
```
adklempner: I don't think this function (or any of the others in this contract) will ever return false since they use `require` to check that the call to `_spender` was successful.

AugustoL: It wont, but it is done like that to follow the same way ERC20 functions are implemented, the `StandardToken` of zeppelin only returns true at the end of the function too.

PR #316, opened 2018-01-30
Update links in README

horyd:
The current ones 404 because it specifies the branch as master, but it works when they are relative to the branch currently being looked at


AugustoL:
thanks @horyd !

PR #317, opened 2018-01-30
Add deploy contracts report

AugustoL:
Add a report of the deploy in production of the smart contracts for the Lif token generation event.


Ratnaram12:
Very good project management ico Token sistem and very nice jobs ico Token

Issue #319, opened 2018-02-15
Make this package releasable on ethpm.com


JirkaChadima:
We should be going for a releasable repository. In this issue I'd like to list and complete all the necessary things that have to be done before the first release is possible. The final product of this issue should be a versioned package on ethpm. Publishing to ethpm should be part of the same process that does #318

**TODO**

- [ ] Cleanup stale branches if there are any
- [ ] Contribution guidelines - https://github.com/blog/1184-contributing-guidelines
- [ ] Up to date documentation in code

The deployed package should contain smart contract source codes.

PR #328, opened 2018-02-21
chore: Puts sol source contracts into artifact published on npmjs

JirkaChadima:
This is a temporary solution until #319 is resolved. It will allow us to simplify dependency management in wt-contracts for now.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,8 +6,6 @@ coverage.json
 allFiredEvents
 scTopics
 package-lock.json
-contracts
-!build/contracts
 migrations
 scripts
 test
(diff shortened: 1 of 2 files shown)


JirkaChadima:
I've added the release commit that is actually now on npmjs... After merging, it would be nice to tag it in this repo as well and cancel the travis build (It should fail anyway in the publish stage as a release with the same name already exists on npm).

JirkaChadima:
@AugustoL I think we need to merge this as well before releasing the Proxy-based token since we still haven't resolved #319

PR #330, opened 2018-04-02
Add solidity-doc generator script with dependencies

AugustoL:
- Add soldmd and solidity-flanneter dependencies.
- Add script to generate documentation based in natspec comments on the contracts.

To generate the documents first the contracts need to be flattened in one file with all the contracts inherited since solmd dont support the imports of zeppelin/solidity module. Once the contracts are flattened in the .flattened directory (ignored by git) the docs are generated form this flattened contracts.

**TODO**

- [x] Review the documentation generated is correct.
- [x] Improve contracts comments and comment all functions.


JirkaChadima:
Am I wrong or the only thing that it actually takes into account are the contract's methods? No other annotation seems to be included in the resulting markdown. That's a disappointment

AugustoL:
@JirkaChadima solcjs is a js binding to use solc with javascript and the other doc-generator projects also use solc.
And solmd only generates the docs with public functions. Is not ideal but this is the generator that requires less out-of-npm dependencies (only solc)

JirkaChadima:
Ok, I was particularly disappointed that it does not generate the whole contract's `@dev` annotation. 

As for the solc: we should add a note to Readme for now, and we also have to deal with it when we will be generating and deploying the documentation automatically. When I have some time, I'll do some PRs into solmd to improve it in a way I like it :)

JirkaChadima:
I've done some PRs in solmd - https://github.com/dpilch/solmd/pull/12, https://github.com/dpilch/solmd/pull/13 to fix at least the basic things. Hopefully, they will be merged soon.

The documentation of the whole contract does not seem to be working on the compiler level :( I really wonder how https://github.com/spalladino/solidity-docgen is doing it

JirkaChadima:
@dpilch Cool, thank you very much! Good luck with your new job

scripts/soldoc.sh
```diff
@@ -0,0 +1,11 @@
+mkdir docs .flattened
+rm docs/*.md
+for f in contracts/*.sol
```
JirkaChadima: We should use something like `for f in $(find contracts -name *.sol)` which would walk through the nested contracts (such as in `deployment`) as well. We could probably ignore contracts we don't want to be generated (such as maybe `LifTokenTest`) as well with find.

docs/LifMarketValidationMechanism.md
```diff
@@ -0,0 +1,390 @@
...
+
+> returns the current buy price at which the MVM offers to buy tokens to burn them
+
+
+
+Outputs
+
+| **type** | **name** | **description** |
+|-|-|-|
+| *uint256* | price | undefined |
```
JirkaChadima: All the outputs can be documented in a nicer way if we use the JSON notation for named outputs, in this case:

```
     @return  { "price": "the current buy price (in eth/lif, multiplied by PRICE_FACTOR)"}
    */
  function getBuyPrice() public view returns (uint256 price) {
```

Such changes in natspec docs will also resolve solmd warnings that occur during `npm run soldoc`

AugustoL: awsm, thx @JirkaChadima

PR #331, opened 2018-04-30
Add Token proxy for LifToken

AugustoL:
Add zos-lib as dependency ans uses his upgradeability proxy contracts to have a new upgradeable version of the token called `LifTokenV0`. The new token version is ERC827.


JirkaChadima:
@augustol Shouldn't this be in the end merged into develop and not master?

AugustoL:
@JirkaChadima you are right, I just changed it to develop.

JirkaChadima:
It looks really good, @AugustoL!

What should be the final state after merging this PR? Should we have the infrastructure code ready for upgrading LifToken at will? Should we have a new LifToken ERC827 compatible version ready? I'm just trying to find out what might be missing in this PR.

AugustoL:
@JirkaChadima the idea is to have all the code deployed on blockchain related with the token here, so anyone can verify it and see what we are using. Once the token is dpeloyed we will add a report of the resources needed to interact and upgrade the token.

JirkaChadima:
Oh, I see you've merged it already @AugustoL. I was thinking that more documentation would be nice. From the first look, it is not really clear what's going on, why we have `LifTokenV0`, which version was used during ICO, which version will we use after Proxy contract is deployed... Maybe README update would be enough. Also we probably need to re-generate the docs folder to reflect the current state.

PR #332, opened 2018-05-15
Fix readme links and add LifTokenV0 description

AugustoL:


JirkaChadima:
It seems a little strange that `LifTokenV0` is titled *The second version* :)

AugustoL:
@JirkaChadima haha, yes, I use that name bacause for upgrading in the smart contracts the contract names are like: `LifTokenV2 is LifTokenV1`. We can change it to LifTokenV1 or LifTokenV2, but I guess since is the first version of the upgradeable token we should use 0 or 1, maybe change LifTokenV0 to LifTokenV1 ?

JirkaChadima:
Could we maybe do something like this?

- `LifTokenV0 is LifToken` (meaning the original instance issued during ICO) - sort of an alias which will not actually ever be used by itself
- `LifTokenV1 is LifToken, ERC827` - the new version with 827 support that we will deploy and migrate to

Also, I've noticed that the LifTokenV0 location is `proxy/test` - the naming sounds weird, why is the directory called test if it's something we actually want to use in production?

AugustoL:
@JirkaChadima good point, I was using test folder because I had it there only for testing. But now it will make sense to remove it form there.

And the naming you proposed sounds better that the ones used, will make thsi changes on another PR

JirkaChadima:
Ok @AugustoL, so what to do with this PR? Merge it as it is, since the docs make sense? And then do another PR with the naming changes we agreed on? We surely have to remember renaming the things in README as well.

AugustoL:
@JirkaChadima sure lets do that, im already working in the renaming and update od the contacts.

PR #333, opened 2018-05-24
Update token proxy contracts

AugustoL:
- Reorder the token contracts.
- Added LifTokenV1 with ERC827 and starting token proxy using LifTokenV0 that is just a LifToken being use right now.
- Added tests for proxy factory contracts.
- Updated zos-lib and truffle dependencies.
- Updated README.md token description and added LifTokenV0 and LifTokenV1 descriptions.


JirkaChadima:
Looks good to me! :clap:

